PR #12, opened 2020-10-07
Translation error in gettext

FransMeerhoff:


R/selectionmodels.R
```diff
@@ -195,7 +195,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
 
 .smFillEstimates           <- function(jaspResults, table, fit, options) {
   
-  overtitleCI <- gettext("95% Confidence Interval")
+  overtitleCI <- gettext("95%% Confidence Interval")
```
TimKDJ, line 198: All of these should be `gettextf` otherwise you get a double `%%` in the result

TimKDJ (changes requested):

R/selectionmodels.R
FransMeerhoff: Thanks! OK

PR #14, opened 2020-10-21
catching up with 0.14 release version

FBartos:
copying changes from the 0.14 release

### DIFF
--- a/R/robustbayesianmetaanalysis.R
+++ b/R/robustbayesianmetaanalysis.R
@@ -135,6 +135,7 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
   "advanced_omit_rhat",
   "advanced_omit",
   "advanced_autofit_error",
+  "advanced_autofit_rhat",
   "advanced_autofit_time_unit",
   "advanced_autofit_time",
   "advanced_autofit",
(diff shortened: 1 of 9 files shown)


TimKDJ:
Travis:
```
test-robustbayesianmetaanalysis.R:1454: failure: Forest plot (Model Averaged) (observed + predicted) matches

Figures don't match: robustbayesianmetaanalysis-forest-plot-model-averaged-prefitted-2.svg
```

FBartos:
fixed, the error was due to the figure being generated with an older version of the package that used incorrect ordering. I also encountered a few more issues when running locally that are fixed now.

Kucharssim (approved):
The code looks ok, the module seems to work (although I did not test all the input options).

@FBartos: Could you link the PRs to stable you are catching up with here? Just for a reference.

FBartos:
It's basically all PRs prior to the 0.14 release. I was in a hurry and didn't have time to add them here as well. Not sure whether the individual PRs are important since neither the Selection Models nor Robust Bayesian Analysis were part of JASP prior to the release.

vandenman:
@FBartos can you rebase and update the figures for ggplot2 version 3.3.2?

FBartos:
ok, I did the rebase and all but one type of plots fails the unit tests for RoBMA. I found that it's due to the ggplot2 3.3.2. Maybe you will have an idea how to fix this @vandenman ?

Basically, I need to construct expressions for y-axis labels programmatically. I did that using bquote and `.()`, however, this seems to throw an error now.

```
library(ggplot2)
df <- data.frame(
  gp = factor(rep(letters[1:3], each = 10)),
  y = rnorm(30)
)

x            <- 5
label_worked <- bquote(~mu ~ "~" ~ .(x))

ggplot(df, aes(gp, y)) +
  geom_point() + 
  scale_y_continuous(
    breaks   = c(0, 1),
    labels   = c(0, label_worked))

```

Interestingly, using `expression()` works, but doesn't use to use variables:

```
label_not_helpful <-  expression(~mu ~ "~" ~ .(x))
ggplot(df, aes(gp, y)) +
  geom_point() + 
  scale_y_continuous(
    breaks   = c(0, 1),
    labels   = c(0, label_not_helpful))
```

vandenman:
Something weird happens here, this might be a bug or regression in ggplot2 itself. In the meantime, this seems to solve the issue:
```r
labels = c(expression(), 0, label_worked)
```
the main difference is that the `c(0, label_worked)` creates a list of expressions whereas `c(expression(), 0, label_worked)` creates one expression with multiple elements. The empty `expression()` is only necessary so that `c()` gives the right behavior, if you really have a constant like 0 you could also do `c(expression(0), label_worked)`.

FBartos:
thanks Don, it worked perfectly. Fix will be out in a new version of the package. Do we merge this prior to that and then create a new PR when the new package version is in JASP (post 0.14,1 I guess?)

vandenman:
@FBartos I think we can just merge this PR in the dynamic module without worrying about 0.14.1. The 0.14.1 could use an older commit of your package (or a separate branch). Not ideal, but hopefully it's only for one release.

TimKDJ:
If you could fix the merge conflict, we can add this

PR #17, opened 2020-11-24
updates to RoBMA 1.2.0.900

FBartos:
- uses "normal" instead of "t" likelihood for model estimation, speeding up the process considerably
- improves the way of dealing with double y-axes
- fixed seed not passed properly
- updates unit tests

The package version required for this to work is not on GitHub yet - the normal likelihood implementation is a side product of larger changes that are not finished and I didn't want to roll those things out publicly yet.
[RoBMA_1.2.0.900.tar.gz](https://github.com/jasp-stats/jaspMetaAnalysis/files/5593181/RoBMA_1.2.0.900.tar.gz)

I'm unable to build JASP since the switch from stable to development and I don't know how to install the new package version if I'm using the model as a development one, but I checked the functionality using unit tests and everything seemed fine.


FBartos:
No need to worry about the 13k lines of code change - these are only the figures for unit tests.

tests/testthat/test-robustbayesianmetaanalysis.R
```diff
@@ -1213,106 +1213,106 @@ fitted_path <- file.path("robmaFit.RDS")
...
+                                        0.164009336133295, 1.00189805889533, "Study 1", 0.62568049934064,
+                                        0.00711458848006568, 905, -0.285009431407208, 0.147833903492011,
+                                        0.145923533777225, 1.00266486701686, "Study 2", 0.562315238784119,
+                                        0.00696160495950572, 888, -0.274416438770882, 0.128132867393647,
+                                        0.127580621830772, 1.00369316402907, "Study 3", 0.524550536497032
+                                   ))
   })
   
-  test_that("Estimated Weights (Ď‰) table results match", {
+  test_that("Estimated Weights (Ï‰) table results match", {
```
vandenman, line 1275: symbols like `Ï‰` will probably make the R CMD CHECK [complain](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Encoding-issues) in the future.

vandenman (approved):
Okay so the changes look fine and I didn't find any bugs in JASP. I left one comment about the tests, but you can ignore that for a PR to stable.

Since you did make many adjustments to the tests, I'd suggest waiting with merging this into the module until we have GitHub actions up and running (hopefully before the end of this week).

One problematic issue I noticed is that you call the progressbar tick function "too often". Basically, every time it's called JASP Engine sends an update with the entire results. My Qt application window is scrolling and showing new json as it looks like new results are received multiple times each second. This makes JASP behave very slowly (e.g., after resizing the jasp application it takes +5 seconds before the scrollbars adjust accordingly).

Although it's good to be aware of this, I'm not entirely sure whether this should be fixed in RoBMA. The progress bar ought to be robust against this heavy usage. @JorisGoosen any ideas?

FBartos:
Thanks for going thru it so quickly!

> One problematic issue I noticed is that you call the progressbar tick function "too often". Basically, every time it's called JASP Engine sends an update with the entire results. My Qt application window is scrolling and showing new json as it looks like new results are received multiple times each second. This makes JASP behave very slowly (e.g., after resizing the jasp application it takes +5 seconds before the scrollbars adjust accordingly).

I don't know how to build this so I can't test the issue but it seems to be weird. The progressbar tick function is called only a when a model is fitted - the default settings fit 12 models, so there should be 12 ticks (the null model is instantaneous, so the first two ticks my arrive basically together), but unless you have only 3 data points with short chains, it should take a second or two to fit them at least.

tests/testthat/test-robustbayesianmetaanalysis.R
FBartos: thanks, I missed this one.

JorisGoosen:
@FBartos maybe you do add results while your fitting a model? Because it does send everything on every change then. Or maybe you made a mistake somewhere. If Don sees that the progressbar is sending too many updates then it must be sending more than 12 right? Because if it is sending so few with such an interval there shouldnt be a problem.

@vandenman The progressbar is a bit of a problem yeah, that is why I opened this issue: https://github.com/jasp-stats/INTERNAL-jasp/issues/1159

It might also be a good idea to change the progressbar interval in stable anyway, Ill do that now.

FBartos:
@JorisGoosen I only add the whole object once it's fitted. And I check and the progress bar tick function is only called once a model is fitted - there is no other usage of it, so I don't know where the additional ticks could've originated.

JorisGoosen:
> @JorisGoosen I only add the whole object once it's fitted. And I check and the progress bar tick function is only called once a model is fitted - there is no other usage of it, so I don't know where the additional ticks could've originated.

Strange, maybe something else is going wrong then. I do have the impression that the webengine has gotten slower lately and was thinking of looking into it soon.

JorisGoosen:
So can this be merged then?

vandenman:
Well, ideally we wait until the unit tests work again. So unless @FBartos wants to also open a PR for other things that depend on this, I'd say we wait a couple days.

vandenman:
The changes looked fine, so the PR to stable can be merged of course

JorisGoosen:
Well, we have to wait for some changes by Koen this weekend anyway. So let's look at this on monday again.

FBartos:
the failed test on MacOS is from Bayesian Meta-Analysis, so this can be merged

vandenman:
could you add a skip to that one figure? Something like
```r
testthat::skip_on_os("mac")
```
or, if you want to convey more information, in the skip:
```r
if (Sys.info()["sysname"] == "Darwin")
  testthat::skip("this test doesn't work on macOS + maybe a reason for future us to look at")
```

FBartos:
done, and the unit tests work now. The falling test was in the Bayesian Meta-Analysis, so it might be relevant for @sophieberkhout .

vandenman:
Great but the part `"this test doesn't work on macOS + maybe a reason for future us to look at"` wasn't meant to be copied literally but to actually fill in that reason :p Anyway I'll merge this now

PR #22, opened 2021-02-15
Restructuring the code and qml interface for incorporation of Cochrane database

FBartos:
the main of this PR is to allow for qml and R code reusability for the future implementation of the Cochrane module
- there is a high overlap in functionality and this will allow us to keep both modules synchronized

fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1187

additional change on EJ's request
- adding a note about the estimation method to the Classical meta-analysis


R/classicalmetaanalysis.R
```diff
@@ -105,1325 +80,3 @@ ClassicalMetaAnalysis <- function(jaspResults, dataset = NULL, options, ...) {
...
-  #   env:         An environement associated with the variables in the formula, see ?as.formula
-  #
-  # Value:
-  #   A formula. See ?formula
-  #
-  terms = sapply(modelTerms, function(x) paste0(unlist(x), collapse = ":"))
-  terms = terms[terms != ""]
-  formula.rhs = paste(terms, collapse = " + ")
-  if (formula.rhs != "") as.formula(paste(" ~ ", formula.rhs), env = env)
-}
```
vandenman, line 1429: I guess these function should not be deleted but moved to classicalmetaanalysiscommon.R instead

FBartos: yes, that was it, thank you!

Kucharssim (approved):
I think it looks ok to me, I don't have an idea how that plays with the Cochrane module, but I believe you thought this through.

I think before we merge, it would be nice if discuss with @sophieberkhout whether to merge this or https://github.com/jasp-stats/jaspMetaAnalysis/pull/28 first, as it they lead to conflicts which one of you will need to resolve.

sophieberkhout:
@FBartos I also don't know how this works with the Cochrane module so I think it would be best if my pull request is merged first (I will have a look today at some changes I still need to make) and you'd have to resolve the conflicts then. Do you agree?

FBartos:
Sorry, but I do disagree. 

Good thing is that you don't really have to worry about the Cochrane module since it imports both qml and R code (it only does some data pre-processing and passes them from an internal database). So, as long as your changes work in the Meta-Analysis module, there is > 95% chance that they won't break Cochrane (if so, I will fix them myself).

Also, this PR mostly divides the main analysis call from some common functions. The only changes to the underlying code are:
- fixing the Classical Meta-Analysis method widged.
- adding an argument to the forest plots and allowing different colors based on a study name prefix,
so, I think that it will be easier to update your specific fixes on top of the split files rather than vice versa.

(also, this PR was here since February, and I mentioned it in the previous JASP Meeting as well as Slack in order to prevent exactly this situation)

sophieberkhout:
Sure, I guess I understand now that it's easier for me to resolve the conflicts. (Also, if you really wanted to prevent this, you could've just let me know personally)

FBartos:
Thanks, I appreciate it. 

Not sure what you meant by that. I really did try to prevent this (kinda surprised that I need to say that explicitly again). I mentioned it here when I started working on it: https://github.com/jasp-stats/INTERNAL-jasp/issues/1187 and I couldn't have known that you were fixing bugs in the Bayesian Meta-Analysis till you made your PR (a month ago), while my PR was being here for more than two months at that point.

Also this message from 16th February:
image

Kucharssim:
Ok, I don't think we need to have a public discussion about who said what on Slack and when as I don't find it appropriate and relevant to this issue. Judging from the fact that this PR is probably ok to be merged now and that both of you seem to agree that it's probably easier for @sophieberkhout to resolve the conflicts than vice versa, I am inclined to merge the current PR first. I hope it's fine @sophieberkhout and that it won't add you too much work on your PR.

If @vandenman does not have any other feedback, then we can merge this.

FBartos:
I just wanted to say that I did not have any intention of causing you additional work and I'm sorry that it resulted in this merge conflict. Also, thank you for resolving the subsequent issue.

PR #23, opened 2021-02-17
updates to RoBMA 1.2.1

FBartos:
Updates the JASP to analysis to use the 1.2.1 version of the package
- it changes the parameterization of RoBMA models accordingly to the changes in our preprint (marginalized random effects)
- as a result, all the numerical results in unit tests change and it's no longer possible to generate true effect estimates (we might bring them back if I figure out how in the future - so I just commented the qml options out)


tests/testthat.R
```diff
@@ -1,4 +1,6 @@
 library(jaspTools)
 library(testthat)
 
+install.packages("RoBMA")
+
```
vandenman, line 5: this is no longer necessary (it's installed from scratch anyway)

tests/testthat/test-robustbayesianmetaanalysis.R
```diff
@@ -630,11 +630,11 @@ fittedPath <- file.path("robmaFit.RDS")
...
-  test_that("Forest plot (Conditional) matches", {
-    plotName <- results[["results"]][["plots"]][["collection"]][["plots_theta"]][["data"]]
-    testPlot <- results[["state"]][["figures"]][[plotName]][["obj"]]
-    expect_equal_plots(testPlot, "forest-plot-conditional-truncated-priors", dir="RobustBayesianMetaAnalysis")
-  })
+#   test_that("Forest plot (Conditional) matches", {
+#     plotName <- results[["results"]][["plots"]][["collection"]][["plots_theta"]][["data"]]
+#     testPlot <- results[["state"]][["figures"]][[plotName]][["obj"]]
+#     expect_equal_plots(testPlot, "forest-plot-conditional-truncated-priors", dir="RobustBayesianMetaAnalysis")
+#   })
```
vandenman, line 637: these and the other commented tests should be skipped instead
```suggestion
   test_that("Forest plot (Conditional) matches", {
      skip(insert a reason for why this fails)
     plotName <- results[["results"]][["plots"]][["collection"]][["plots_theta"]][["data"]]
     testPlot <- results[["state"]][["figures"]][[plotName]][["obj"]]
     expect_equal_plots(testPlot, "forest-plot-conditional-truncated-priors", dir="RobustBayesianMetaAnalysis")
   })
```

tests/testthat/test-robustbayesianmetaanalysis.R
```diff
@@ -1067,7 +1067,7 @@ fittedPath <- file.path("robmaFit.RDS")
   options$resultsIndividualSingleNumber <- 12
   options$resultsModelsBF <- "inclusion"
   options$resultsModelsOrder <- "default"
-  options$resultsTheta <- TRUE
+  options$resultsTheta <- FALSE
```
vandenman, line 1070: why? what changed? can we just temporarily skip the failing tests?

FBartos: The package uses different parametrization of the likelihood so it does not estimate the Theta coefficients anymore. I kept some functionality associated with their plotting and etc in the JASPs backend I manage to add them back but they are not available to users anymore and the error would throw an error if they were requested.

tests/testthat/test-robustbayesianmetaanalysis.R
FBartos: fixed

PR #28, opened 2021-04-15
Fixes Bayesian Meta-Analysis

sophieberkhout:
- Fixed wrong legend label and wrong table label ([#1236](https://github.com/jasp-stats/jasp-issues/issues/1236))
- I fixed some labels in the BF sequential plot
- The last result of the sequential analysis is now taken from the state (removing small differences and making it slightly more efficient)
- I made drawing the prior optional in the cumulative forest plot (request from EJ a while ago)

PR #38, opened 2021-06-02
Fix Bayesian Meta-Analysis 2

sophieberkhout:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1236
New pull request (old one: #28) because there were merge conflicts with #22.

Regarding one of the comments in #28 about this plot: 
image

I kept it like this. It is not possible to show `RE H1` because it is too long and the text will be cut off. I think `r H1` would not be too confusing, since the BF is also labeled with just the `r` and the `f`.

### DIFF
--- a/R/bayesianmetaanalysiscommon.R
+++ b/R/bayesianmetaanalysiscommon.R
@@ -1236,25 +1236,26 @@
   } 
   
   # Scale the height and width of the plot
-  height <- heightCum <- 100 + nrow(dataset) * 30 
-  if(options$forestPlot == "plotForestBoth") {
-    height <- 100 + nrow(dataset) * 30 + 50 + nrow(dataset) * 30
-  } 
+  heightCum <- 100 + nrow(dataset) * 30 
   width  <- 500 + (nchar(max(studyLabels)) * 5)
   
   # Create empty plot
   if(is.null(forestContainer[["forestPlot"]]) && options$checkForestPlot) {
     
-    # title of plot based on observed/estimated 
-    if(options$forestPlot == "plotForestObserved" || options$modelSpecification == "FE"){
-      title <- gettext("Observed study effects")
+    # title and height of plot based on observed/estimated 
+    if(options$forestPlot == "plotForestObserved"){
+      title  <- gettext("Observed study effects")
+      height <- 100 + nrow(dataset) * 30
     } else if(options$forestPlot == "plotForestEstimated"){
-      title <- gettext("Estimated study effects")
+      title  <- gettext("Estimated study effects")
+      height <- 100 + nrow(dataset) * 30
     } else if(options$forestPlot == "plotForestBoth"){
       title  <- gettext("Observed and estimated study effects")
+      height <- 150 + 2 * (nrow(dataset) * 30)
     }
     
     forestPlot <- createJaspPlot(plot = NULL, title = title, height = height, width = width)
+    
     # Fill plot
     forestPlot$dependOn(c("forestPlot", "checkForestPlot", 
                           "orderForest", "forestPlotOrder", "showLabels"))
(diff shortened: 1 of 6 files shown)


R/bayesianmetaanalysiscommon.R
```diff
@@ -415,11 +398,63 @@
...
+  
+  d                       <- .bmaPriors(jaspResults, options)[["d"]]
+  # Fix voor truncated priors
+  priorSamples            <- sample(seq(-10, 10, by = 0.0001), size = 1e6, replace = TRUE, prob = d(seq(-10, 10, by = 0.0001)))
+  seqResults$mean[1]      <- mean(priorSamples)
+  seqResults$lowerMain[1] <- quantile(priorSamples, probs = 0.025)
+  seqResults$upperMain[1] <- quantile(priorSamples, probs = 0.975)
+  
+  # meta analysis cannot run with only 1 study so it starts with 2
+  # the final result is already in the state, so we do not have to run it again (n-1)
```
Kucharssim, line 446: nice comment!

R/bayesianmetaanalysiscommon.R
```diff
@@ -1185,51 +1220,54 @@
...
-    title <- gettext("Observed study effects")
-  }
-  
   # Check if ready    
   if(!ready){
     return()
   } 
   
+  # Scale the height and width of the plot
+  heightCum <- 100 + nrow(dataset) * 30
```
Kucharssim, line 1239: I personally prefer to call variable names like `heightCumulative` or `totalHeight` - it's only a couple of extra key strokes and adds clarity to the code. But it's your choice of course whether to change it...

R/bayesianmetaanalysiscommon.R
```diff
@@ -1546,12 +1584,20 @@
   
   df <- data.frame(effectSize = meanMain, studyLabels = studyLabels, y = length(meanMain):1)
   
+  if(!options$addPrior) {
+    idx <- which(df$studyLabels == "Prior")
```
Kucharssim, line 1588: Nice solution to the indexing from before!

Kucharssim (commented):
It seems ok to me, the issues are still fixed and the weird QML issue is gone!

I just have a question regarding this:
>I kept it like this. It is not possible to show RE H1 because it is too long and the text will be cut off. I think r H1 would not be too confusing, since the BF is also labeled with just the r and the f.

Can't we name it H<sub>F</sub> and H<sub>R</sub> or something like that (then BF<sub>FR</sub> makes also more sense)? Also, the legend of the plot shows "Prior H<sub>1</sub>" only, but in your case it should then be called "Prior Random H<sub>1</sub>" I think. Or if you embrace my suggestion, just "Prior H<sub>R</sub>". If we resolve this I am happy to merge this :)

R/bayesianmetaanalysiscommon.R
sophieberkhout: I changed it to `heightCumulative` :)

sophieberkhout:
> Can't we name it HF and HR or something like that (then BFFR makes also more sense)? Also, the legend of the plot shows "Prior H1" only, but in your case it should then be called "Prior Random H1" I think. Or if you embrace my suggestion, just "Prior HR". If we resolve this I am happy to merge this :)

I discussed this with EJ and Quentin and changed a couple of labels accordingly. We chose to use Hr1 and Hf1 (no subscript since the regular H0 and H1 also get no subscripts in the pizza plot in JASP). But then we also had to change the BF labels to BF<sub>_r1f1_</sub>. And also for the constrained random effects analysis. 
Then, it also needed to be changed accordingly for the sequential plot but only for the RE and CRE analyses, as the model-averaged sequential does not need the extra 1 subscript (very complicated, I know haha). 

Regarding the prior, this is the prior for the random effects as well as the model-averaged model. However, unfortunately, there is no posterior for the model-averaged heterogeneity yet.

I also fixed a bug in the table notes for the constrained random effects analysis.

Kucharssim (approved):
Ok, this looks good now!

PR #43, opened 2021-06-24
Update unit tests to use vdiffr 1.0

sophieberkhout:
For https://github.com/jasp-stats/INTERNAL-jasp/issues/1408


sophieberkhout:
Oops there was a commit from Frantisek that should not have been here, so I reverted it. Then this should not have conflicts with my other pull request I think.

Kucharssim:
> Oops there was a commit from Frantisek that should not have been here

The safest option to avoid these problems is to always make sure to rebase on master. https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/git-guide.md provides a guide how to do that (on `jasp-desktop/development`, but it works the same way for any module and branch.)

TimKDJ (approved):
I'll squash the commits, but for future reference to clean up the current situation you could do

```
git reset head~3
git add .
git commit -m "update meta analysis to vdiffr 1.0"
git push --force
```

sophieberkhout:
Thanks Simon and Tim, I will make sure to do this next time!

PR #35, opened 2021-05-26
RoBMA - fixes

FBartos:
fixes: jasp-stats/jasp-test-release#1277
fixes: jasp-stats/jasp-test-release#1273
fixes: jasp-stats/jasp-test-release#1279
fixes: jasp-stats/jasp-test-release#1284

1277, 1279, and 1284 are fixed in the new version of the package RoBMA 2.0 that's about to be released. However, it breaks backward compatibility and won't be ready for this JASP release.

Some of the fixes would be actually pretty easy to do, like 1279 (where a wrong argument is checked internally) and 1284 (where an x_axis label is not properly generated), however, the old version does not compile on R4.0 and making it compile would result in breaking much more code (therefore the new upcoming version).

There are unfortunately no walk arounds, so I have hidden broken options from the user in the qml interface.

PR #52, opened 2021-07-19
adding WAAP-WLS & PET-PEESE + small improvements to selection models

FBartos:
- adds WAAP-WLS & PET-PEESE analyses
- improves documentation in selection models
- fixes the standard errors of transformed effect size estimates not being transformed to the original scale in cases (in cases when correlations were supplied as input)

Also separates the transformations from the selection models, so they can be shared with the WAAP-WLS & PET-PEESE.
 
(RoBMA unit tests will fail - need the https://github.com/jasp-stats/jaspMetaAnalysis/pull/35 to me merged)

I think that we need better handling of effect size transformations across the whole module. I can imagine having two separate analyses. One dedicated to **effect size computations**, e.g., from the sample sizes, means, sds, ..., and another one dedicated to **effect size transformations** between different effect size estimates (I have already implemented convenience functions for this in RoBMA 2.0). Together, the effect size computation & transformations could be handled outside of the individual analyses. Then, we could easily add a dropdown specifying what kind of input did the user choose and whether he wants the out to be transformed to a different standardized scale - it would simplify the input menus, especially in the modules where I implemented the possibility to supply correlations or/and log(OR) as input.
(I will copy this comment to a separate issue).

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -1,5 +1,6 @@
 import(jaspBase)
 export(ClassicalMetaAnalysis)
 export(SelectionModels)
+export(WaapWlsPetPeese)
 export(BayesianMetaAnalysis)
 export(RobustBayesianMetaAnalysis)
--- a/R/selectionmodels.R
+++ b/R/selectionmodels.R
@@ -16,24 +16,24 @@
 #
 
 SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
-  
+
   if (.smCheckReady(options)) {
     # get the data
     dataset <- .smGetData(dataset, options)
     dataset <- .smCheckData(jaspResults, dataset, options)
-    
+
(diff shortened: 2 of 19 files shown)


sophieberkhout:
- I thought that the axes labels of plots only had the first word capitalized in JASP, but now that I am going through some analyses, there seems to be no consistency in this. So I guess you can disregard this.
- There is a strange error message when the 'Multiplicative heterogeneity estimates' table is selected and then you click on the 'Correlations & N' radiobutton and it doesn't go away if you go back to 'Effect size & SE'.
image


Other then that, it looks good to me! The help files are very helpful also :)

sophieberkhout (approved):
The unit tests fail because of the robust meta analysis, so this pull request is fine I believe.

PR #56, opened 2021-07-23
Revert "adding WAAP-WLS & PET-PEESE + small improvements to selection models"

AlexanderLyNL:
Reverts jasp-stats/jaspMetaAnalysis#52

I revert this, because of the feature stop for the release.


AlexanderLyNL:
This shouldn't have been merged because of the failed unit test either. This means that the unit is dodgy/unreliable, which defeats the purpose of having them to begin with.

PR #64, opened 2021-08-31
Bayesian Meta-Analysis - Fix plot labels & enable advanced settings CRE & fix sequential plot

sophieberkhout:
This pull request adjusts the labels for sequential Bayes factor plots and posterior model probabilities plot for the constrained random effects model, see [#1527](https://github.com/jasp-stats/jasp-test-release/issues/1527).

Moreover, I enabled the advanced settings for this model as they finally work (there was a problem with the `metaBMA` package before). (this should also make the reviewing easier, as you can now decrease the number of iterations and chains to make it go faster)

Also, I fixed an issue with the posterior model probabilities plot where the probabilities for when there is only 1 study should be the same as the prior probabilities (same as with 0 studies) as the analysis does not work with only 1 study.

### DIFF
--- a/R/bayesianmetaanalysiscommon.R
+++ b/R/bayesianmetaanalysiscommon.R
@@ -54,7 +54,7 @@
 }
 
 .bmaDependencies <- c("effectSize", "standardError", "confidenceInterval", "modelSpecification",
-                      "allPos", "allNeg", "priorH0FE", "priorH1FE", "priorH0RE", "priorH1RE",
+                      "allPos", "allNeg", "priorH0FE", "priorH1FE", "priorH0RE", "priorH1RE", "priorH1CRE",
                       "priorES", "informativeCauchyLocation", "informativeCauchyScale",
                       "checkLowerPrior", "checkUpperPrior", "lowerTrunc", "upperTrunc",
                       "informativeNormalMean", "informativeNormalStd",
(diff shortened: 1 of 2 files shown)


vandenman:
@sophieberkhout are there any open test issues related to this PR? That kinda determines wether or not we should merge this for 0.15.

Either way, there is quite some funky stuff going on here (e.g., `jaspGraphs:::` and `domain="R-jaspGraphs"`), ideally, we'd export these functions and avoid having you do `domain="R-jaspGraphs"` somehow (which would be an argument for postponing this to 0.16).

sophieberkhout:
@vandenman well, the test issue [#1527](https://github.com/jasp-stats/jasp-test-release/issues/1527) was closed by EJ after I showed him the new plot labels from this PR. Besides that there are no open issues.

I think it would be nice to have this in 0.15 as labels now indicate the wrong model (random instead of ordered) but I think this analysis (constrained random effects) is not used often so maybe it can wait?

sophieberkhout:
@vandenman would you still be able to look at this?

vandenman:
I looked at it, but the problem is that `jaspGraphs:::fixTranslationForExpression` is actually quite broken. I'm looking at a way to use HTML tags for the subscripts. Until then, I would postpone this.

sophieberkhout:
Ah okay thanks :)

AlexanderLyNL:
@sophieberkhout could you please check and fix the conflicts?

FBartos:
@sophieberkhout, do you remember whether these issues are still a problem or have been fixed already?

vandenman:
closed for now, feel free to reopen if this is still necessary.

PR #65, opened 2021-08-31
Bayesian Meta-Analysis - bug fix prior model probabilities

sophieberkhout:
This pull request fixes the issue [#1514](https://github.com/jasp-stats/INTERNAL-jasp/issues/1514). It would be best if this can be merged before the release.

I had to basically revert some of the changes @FBartos made in two QML files. Can you check if this structure also works for you?

(this pull request conflicts with my other one (#64), therefore I will have to adjust that one after this is merged)

### DIFF
--- a/inst/qml/qml_components/BayesianMetaAnalysisAdvanced.qml
+++ b/inst/qml/qml_components/BayesianMetaAnalysisAdvanced.qml
@@ -24,11 +24,6 @@ Section
 {
 	columns: 	2
 	title: 		qsTr("Advanced")
-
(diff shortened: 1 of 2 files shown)


FBartos:
I will check it over the weekend.

FBartos:
It looks good and works. 

Maybe for the next release but we might want to show the prior model probabilities even before fitting the models?
image

FBartos (approved):

sophieberkhout:
Thanks! And I was thinking the same thing, I will add that for the next release :)

PR #75, opened 2021-09-24
Adding Meta-Analysis of Prediction Model Performance (Utrecht project)

FBartos:
Adds classical/Bayesian meta-analysis of prediction model performance based on the metamisc package

I would recommend [EuroSCORE.csv](https://github.com/jasp-stats/jaspMetaAnalysis/files/7223782/EuroSCORE.csv) data set for testing since the analysis requires a combination of observed/expected number of events, number of participants, or c-statistics + se / CI, all of which are bounded to be positive.

known shortcomings:
- unit tests for the Bayesian version are disabled due to a missing set.seed functionality in metamisc 
- missing documentation, will be supplied later by author of the package


vandenman:
@FBartos did you mean to assign a reviewer?

FBartos:
@vandenman Yes, I need someone to review this. Would you be willing to take a look?

tests/testthat/test-bayesianpredictionperformance.R
```diff
@@ -0,0 +1,425 @@
...
+  options$priorTauTMax <- 10
+  options$priorTauTMin <- 0
+  options$priorTauTScale <- 1.5
+  options$priorTauUMax <- 2
+  options$priorTauUMin <- 0
+  options$sample <- 5000
+  set.seed(1)
+  dataset <- structure(list(Study = structure(c(14L, 2L, 8L, 7L, 9L, 5L, 12L,
+                                                11L, 10L, 20L, 3L, 18L, 19L, 21L, 13L, 20L, 15L, 1L, 1L, 17L,
+                                                6L, 4L, 16L), .Label = c("Barilli", "Biancari", "Borde", "Borracci",
```
vandenman, line 187: Add a CSV file to tests and load it? Same for all the others.

tests/testthat/test-bayesianpredictionperformance.R
```diff
@@ -0,0 +1,425 @@
...
+  test_that("Observed-Expected Ratio Meta-Analysis Summary table results match", {
+    table <- results[["results"]][["summaryTable"]][["data"]]
+    jaspTools::expect_equal_tables(table,
+                                   list(1.09088086726799, 0.88575857156341, 0.285932954156389, 1.34586691302466,
+                                        2.45439237587575))
+  })
+}
+
+# O:E: Poisson/log (+ asymmetry tests)
+{
```
vandenman, line 137: This does nothing, so please remove it everywhere.

tests/testthat/test-bayesianpredictionperformance.R
```diff
@@ -0,0 +1,425 @@
...
+  options$diagnosticsAcPlot <- TRUE
+  options$diagnosticsGelmanRubinPlot <- TRUE
+  options$diagnosticsRmPlot <- TRUE
+  options$exportColumns <- FALSE
+  options$exportCstat <- "JaspColumn_17_Encoded"
+  options$exportCstatlCI <- "JaspColumn_18_Encoded"
+  options$exportCstatuCI <- "JaspColumn_19_Encoded"
+  options$exportOE <- "JaspColumn_20_Encoded"
+  options$exportOElCI <- "JaspColumn_21_Encoded"
+  options$exportOEuCI <- "JaspColumn_22_Encoded"
```
vandenman, line 17: ... what? The data doesn't use these as column names? Can't you use `Study` and `c.index` directly?

inst/qml/qml_components/PredictionPerformancePriors.qml
```diff
@@ -0,0 +1,145 @@
...
+
+	Group
+	{
+
+		title: 				if(measure == "OE"){qsTr("Summary estimate (O:E Ratio)")}else{qsTr("Summary estimate (C statistic)")}
+		columns:			2
+
+			DoubleField
+			{
+				label: 			qsTr("Normal   mean:")
```
vandenman, line 40:
```suggestion
				label: 			qsTr("Normal mean:")
```

inst/qml/qml_components/PredictionPerformancePriors.qml
```diff
@@ -0,0 +1,145 @@
...
+	title:		qsTr("Priors")
+	expanded:	false
+	columns:	1
+	
+	property string	measure:			"OE"
+
+	Group
+	{
+
+		title: 				if(measure == "OE"){qsTr("Summary estimate (O:E Ratio)")}else{qsTr("Summary estimate (C statistic)")}
```
vandenman, line 35: if you want, you can also use a ternary here:
```suggestion
		title: 				measure === "OE" ? qsTr("Summary estimate (O:E Ratio)") :  qsTr("Summary estimate (C statistic)")
```
`===` instead of `==` might be a bit more appropriate although it doesn't really matter.

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,259 @@
...
+BayesianPredictionPerformance  <- function(jaspResults, dataset, options, state = NULL) {
+
+  options[["method"]] <- "BAYES"
+  ready <- .metamiscReady(options)
+
+  if (ready)
+    dataset <- .metamiscGetData(options, dataset)
+
+  if (ready)
+    .metamiscFitModelBayesian(jaspResults, options, dataset)
```
vandenman, line 27:
```suggestion
  if (ready) {
    dataset <- .metamiscGetData(options, dataset)
    .metamiscFitModelBayesian(jaspResults, options, dataset)
  }
```

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,259 @@
...
+    adapt      = options[["adapt"]],
+    burnin     = options[["burnin"]],
+    sample     = options[["sample"]],
+    n.chains   = options[["chains"]]
+  ),error = function(e)e)
+
+  # error handling
+  if (any(class(fit) %in% c("simpleError", "error"))) {
+    .quitAnalysis(paste0("metamisc package failed with the following error: '", fit[["message"]], "'"))
+  }
```
vandenman, line 120: So this is not ideal behavior for the user. Better is to do
```r
container$setError(gettextf("metamisc package failed with the following error: '%s''", fit[["message"]])
```
where `container` is the outermost container of the results.
Also, this cannot be translated, of course, you need to use `gettextf`.

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,259 @@
...
+    return()
+
+  p <- try(metamisc::rmplot(jaspResults[["model"]][["object"]]))
+
+  if (isTryError(p)) {
+    rmPlot$setError(.extractErrorMessage(p))
+    return()
+  }
+
+  p <- jaspGraphs::themeJasp(p, sides = "bl") + ggplot2::xlab(gettext("Iteration")) + ggplot2::ylab(gettext("Running mean"))
```
vandenman, line 177: `themeJasp` is deprecated in favor of explicitly using `geom_rangeframe` and `themeJaspRaw`.
```suggestion
  p <- p + 
    jaspGraphs::geom_rangeframe() +
    jaspGraphs::themeJaspRaw() + 
    ggplot2::xlab(gettext("Iteration")) + ggplot2::ylab(gettext("Running mean"))
```

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,259 @@
...
+  if (options[["inputMeasure"]] != "" && sum(unlist(options[["inputCI"]]) != "") == 2)
+    dataset$computable <- ifelse(!is.na(dataset[,options[["inputMeasure"]]]) & !is.na(dataset[, options[["inputCI"]][[1]][1]]) & !is.na(dataset[, options[["inputCI"]][[1]][2]]), TRUE, dataset$computable)
+
+  if (options[["measure"]] == "cstat" && options[["inputMeasure"]] != "" && options[["inputN"]] != "" && options[["inputO"]] != "")
+    dataset$computable <- ifelse(!is.na(dataset[, options[["inputMeasure"]]]) & !is.na(dataset[, options[["inputN"]]]) & !is.na(dataset[, options[["inputO"]]]), TRUE, dataset$computable)
+
+  if (options[["measure"]] == "OE" && options[["inputE"]] != "" && options[["inputO"]] != "")
+    dataset$computable <- ifelse(!is.na(dataset[, options[["inputE"]]]) & !is.na(dataset[, options[["inputO"]]]), TRUE, dataset$computable)
+
+  dataset <- dataset[dataset$computable, ]
```
vandenman, line 256: at the end here you'll always have dataset$computable in there? why not:

```suggestion
  
  computable <- rep(FALSE, nrow(dataset))

  if (options[["inputMeasure"]] != "" && options[["inputSE"]] != "")
    computable <- computable | (!is.na(dataset[, options[["inputMeasure"]]]) & !is.na(dataset[, options[["inputSE"]]])

  if (options[["inputMeasure"]] != "" && sum(unlist(options[["inputCI"]]) != "") == 2)
    computable <- computable | (!is.na(dataset[,options[["inputMeasure"]]]) & !is.na(dataset[, options[["inputCI"]][[1]][1]]) & !is.na(dataset[, options[["inputCI"]][[1]][2]])

  if (options[["measure"]] == "cstat" && options[["inputMeasure"]] != "" && options[["inputN"]] != "" && options[["inputO"]] != "")
    computable <- computable | (!is.na(dataset[, options[["inputMeasure"]]]) & !is.na(dataset[, options[["inputN"]]]) & !is.na(dataset[, options[["inputO"]]]))

  if (options[["measure"]] == "OE" && options[["inputE"]] != "" && options[["inputO"]] != "")
    computable <- computable | (!is.na(dataset[, options[["inputE"]]]) & !is.na(dataset[, options[["inputO"]]]))

  dataset <- dataset[computable, ]
```
so basically, join the logicals with `computable | (conditions)` rather than `ifelse(condition, TRUE, computable)`.

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,628 @@
...
+
+ClassicalPredictionPerformance   <- function(jaspResults, dataset, options, state = NULL) {
+
+  ready <- .metamiscReady(options)
+
+  if (ready)
+    dataset <- .metamiscGetData(options, dataset)
+
+  if (ready)
+    .metamiscFitModel(jaspResults, options, dataset)
```
vandenman, line 27:
```suggestion
  if (ready) {
    dataset <- .metamiscGetData(options, dataset)
    .metamiscFitModel(jaspResults, options, dataset)
  }
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,633 @@
...
+    return()
+
+  p <- plot(jaspResults[["model"]][["object"]])
+
+  if (isTryError(p)) {
+    forestPlot$setError(.extractErrorMessage(p))
+    return()
+  }
+
+  p <- jaspGraphs::themeJasp(p, sides = "b") +
```
vandenman, line 228: same thing about `themeJaspRaw` and `geom_rangeframe`.

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,633 @@
...
+    funnelTestPlots <- createJaspContainer(title = gettext("Funnel Plot Asymmetry Plots"))
+    funnelTestPlots$dependOn(c(.metamiscDependencies, "funnelAsymmetryTestPlot", "funnelAsymmetryTest", if (options[["method"]] == "BAYES") .metamiscDependenciesBayesian))
+    funnelTestPlots$position <- 4
+    jaspResults[["funnelTestPlots"]] <- funnelTestPlots
+  }
+
+  fatFits <- jaspResults[["modelsFat"]]$object
+
+  for(i in seq_along(fatFits)) {
+    if (!any(class(fatFits[[i]]) %in% c("simpleError", "error")) && is.null(funnelTestPlots[[.metamiscFitFunnelAsymmetryNames(fatFits[[i]]$method)]])) {
```
vandenman, line 438: could you use `inherits` here?

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,633 @@
...
+      tempFunnelPlot$position <- i
+      tempFunnelPlot$dependOn(.metamiscFitFunnelAsymmetryOptions(fatFits[[i]]$method))
+      funnelTestPlots[[.metamiscFitFunnelAsymmetryNames(fatFits[[i]]$method)]] <- tempFunnelPlot
+
+      tempPlot <- tryCatch(.metamiscFitFunnelAsymmetryggPlot(fatFits[[i]]), error = function(e)e)
+
+      if (any(class(tempPlot) %in% c("simpleError", "error"))) {
+        tempFunnelPlot$setError(tempPlot$message)
+      } else{
+        tempPlot <- jaspGraphs::themeJasp(tempPlot)
```
vandenman, line 453: same about `themeJaspRaw`.

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,633 @@
...
+    stop("Argument 'x' must be an object of class \"fat\".")
+  if (confint.level < 0 | confint.level > 1) {
+    stop("Argument 'confint.level' must be between 0 and 1.")
+  }
+
+  xval <- x$model$data[, "y"]
+  if (x$method %in% c("E-UW", "E-FIV")) {
+    ylab <- "Standard error"
+    yval <- (x$model$data[, "x"])
+    ylim <- rev(c(0, max(yval, na.rm = T)))
```
vandenman, line 497: `T` -> `TRUE`.

vandenman (changes requested):
Here is some superficial feedback. I'll try out the code in JASP after you fix the comments (just request another review when you're ready).

tests/testthat/test-bayesianpredictionperformance.R
FBartos: how about loading it with ```data("EuroSCORE", package = "metamisc")``` ?

tests/testthat/test-bayesianpredictionperformance.R
FBartos: changed to wrapping by ----

tests/testthat/test-bayesianpredictionperformance.R
FBartos: changed to wrapping by ----

R/bayesianpredictionperformance.R
FBartos: hmm, but when the fitting function fails, the rest of the analysis is unusable, so isn't it better to completely stop it?

inst/qml/qml_components/PredictionPerformancePriors.qml
FBartos: ok, and propagated

R/bayesianpredictionperformance.R
FBartos: ok, also propagated to the rest

R/classicalpredictionperformance.R
FBartos: done

R/classicalpredictionperformance.R
FBartos: yeah, I just copied the code from the package to update the base plot functions to ggplot. Should be eventually taken care of withing the package itself

R/classicalpredictionperformance.R
FBartos: sure, done

R/bayesianpredictionperformance.R
vandenman: Well, the results will look pretty similar, except that you can maybe still show descriptive information. However, if users click on some options while there is an error they would get feedback in the form of greyed-out tables/ plots.

With this change the other functions become:
```R
table <- createJaspTable(...) # or some plot

table$addColumnInfo(...)

container[[name]] <- table

if (!ready || container$getError()) # <- should be the only change necessary to the other functions
  return()
```

FBartos: ok, changed

vandenman:
Some initial things I found:

- [ ] As soon as you start either the frequentist or the Bayesian analysis, empty new columns are added to the data: 
image

- [ ] If you add `Unicode` under Study labels you get a weird error message about negative numbers:
image

- [ ] The funnel plot asymmetry tests seem a little bit buggy:
image
  - [ ] The table (red) does not show any values, but there is no footnote.
  - [ ] Two plots (dark red) cannot be computed without any reasonable error message.
  - [ ] The Debray plot (yellow) has a funky y-axis.

- [ ] If some data must be entered before a particular statistic can be computed, then perhaps make the option greyed out when e.g., the number of events is not given.
image

- [ ] The UI is too wide for the priors.
image

- [ ] Under MCMC Diagnostics:
image
  - [ ] The title of the Gelman Rubin plot is Autocorrelations plot
  - [ ] For all MCMC diagnostic plots there are four panels that each have a title, except for the Gelman Rubin plot which has only 2 panels.
      - [ ] What do these panels mean? Could these also have 4 panels, one for each chain? Or at least a title?
  - [ ] In the Gelman Rubin plot, it is unclear what the different line colors represent (mean and xx ci?). Maybe add a legend to clarify?

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,269 @@
...
+
+  if (ready) {
+    dataset <- .metamiscGetData(options, dataset)
+    .metamiscFitModelBayesian(jaspResults, options, dataset)
+  }
+
+  .metamiscSummaryTable(jaspResults, options)
+
+  if (jaspResults[["summaryTable"]]$getError())
+    return()
```
vandenman, line 31: This is not what I meant about the error handling, it is meaningless to do return here.

R/bayesianpredictionperformance.R
```diff
@@ -0,0 +1,269 @@
...
+  imgHeight  <- 1000
+  imgWidth   <- 520
+
+  priorAndPosteriorPlot   <- createJaspPlot(title = gettext("Prior and posterior plot"), width = imgWidth, height = imgHeight)
+  priorAndPosteriorPlot$position <- 2.1
+  priorAndPosteriorPlot$dependOn(c("priorAndPosteriorPlot", c(.metamiscDependencies, .metamiscDependenciesBayesian)))
+  jaspResults[["priorAndPosteriorPlot"]] <- priorAndPosteriorPlot
+
+  if (!ready)
+    return()
```
vandenman, line 142: instead, you should change this to
```suggestion
  if (!ready || jaspResults[["summaryTable"]]$getError())
    return()
```
for all output elements.

vandenman (changes requested):

R/bayesianpredictionperformance.R
FBartos: I do not really see the difference between these two solutions apart from the fact that I will need to write ```jaspResults[["summaryTable"]]$getError()``` 4 more times in classical and 8 more times in Bayesian version. If the model can't be fit, the remaining options should become unavailable - both cases make that happen.

vandenman: Here's an example. Without this, no matter what options you select, once there is an error this is what the user gets:

image

In other words, once there is an error it looks like the analysis is completely unresponsive. That is a bad user experience.

If you do as I suggest, a user will still see empty output elements are still whenever they select the corresponding options:

image

Had you put the error on jaspResults (or any common container), then the empty plot element would also be shown greyed out like the table, instead of black.

See also https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/r-analyses-guide.md#propagating-errors-through-a-container

FBartos:
> As soon as you start either the frequentist or the Bayesian analysis, empty new columns are added to the data.

I know, it's not ideal but I discussed this with Joris and this is the expected behavior when using ```ComputedColumnField``` that I have been advised to use.


> If you add Unicode under Study labels you get a weird error message about negative numbers:

It does not happen to me on Windows but the labels should be omitted from the subsequent checks now.


> The funnel plot asymmetry tests seem a little bit buggy:

Yeah, the e


> If some data must be entered before a particular statistic can be computed, then perhaps make the option greyed out when e.g., the number of events is not given.

Thanks, that's a much better way of dealing with the issue indeed.


> The UI is too wide for the priors.

I see, that's a UI scaling issue (I did not have the problem even at 90% scaling previously though).


> Under MCMC Diagnostics:
> The title of the Gelman Rubin plot is Autocorrelations plot

Thanks, fixed.

> For all MCMC diagnostic plots there are four panels that each have a title, except for the Gelman Rubin plot which has only 2 panels.

I already suggested these things to Thomas (author of the package that we are implementing the analysis for) and he might change those plots in the future. They are a result of the packages used to create them but they might be re-written customly in future releases. 

> In the Gelman Rubin plot, it is unclear what the different line colors represent (mean and xx ci?). Maybe add a legend to clarify?

fixed

R/bayesianpredictionperformance.R
FBartos: ok, I see the point now. I didn't expect that we want to show empty figures when the whole analysis is crashed. I changed it appropriately now.

FBartos:
So, I also updated the error handling for funnel plots now. I did not fix the bad ticks on y-axis for this one specific figure though - I copied code for the figure from the package and this is how it was supposed to be created in base plot (IMO, this is just an extreme edge case that could be handled with the now existing plot editing?)

To sum this and the above post. I think I fixed everything apart from:
- the automatically added columns (reason in post above)
- edge case y-axis scaling for funnel plots
- (unnecessary) multipanel MCMC diagnostics

vandenman:
> I copied code for the figure from the package and this is how it was supposed to be created in base plot (IMO, this is just an extreme edge case that could be handled with the now existing plot editing?)

Plot editing is meant for tweaking figures when we cannot automatically produce the right figure.

The goal is not that we lower the bar for plots and show things like these as default output with the argument "the user can fix it themselves if they want to":
image

Please just fix this.


>     For all MCMC diagnostic plots there are four panels that each have a title, except for the Gelman Rubin plot which has only 2 panels.
> 
> I already suggested these things to Thomas (author of the package that we are implementing the analysis for) and he might change those plots in the future. They are a result of the packages used to create them but they might be re-written customly in future releases.

So they come straight out of the package right now? In that case, just leave it as is.

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+  } else if (x$method == "D-FAV") {
+    ylab <- "Total events"
+    yval <- (x$model$data[, "x"])
+    ylim <- rev(c(0, max(yval, na.rm = TRUE)))
+    xlim <- c(min(c(0, xval)), max(xval))
+    step <- ((max(yval) - min(yval))/4)
+    yax  <- c(plyr::round_any(1/min(yval),10^(sapply(round(1/min(yval)), nchar) - 1)),
+              plyr::round_any(1/seq(step, 4 * step, by = step), 10), plyr::round_any(1/max(yval), 10))
+  } else {
+    stop("Plot not supported!")
```
vandenman, line 550:
```suggestion
    stop("Plot not supported!", domain = NA)
```
otherwise, people will actually translate this.

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+
+  if (!inherits(x, "fat"))
+    stop("Argument 'x' must be an object of class \"fat\".")
+  if (confint.level < 0 | confint.level > 1) {
+    stop("Argument 'confint.level' must be between 0 and 1.")
+  }
+
+  xval <- x$model$data[, "y"]
+  if (x$method %in% c("E-UW", "E-FIV")) {
+    ylab <- "Standard error"
```
vandenman, line 516:
```suggestion
    ylab <- gettext("Standard error")
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    yval <- (x$model$data[, "x"])
+    ylim <- rev(c(0, max(yval, na.rm = TRUE)))
+    xlim <- c(min(c(0, xval)), max(xval))
+  } else if (x$method %in% c("M-FIV")) {
+    ylab <- "Sample size"
```
vandenman, line 521:
```suggestion
    ylab <- gettext("Sample size")
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    yval <- (x$model$data[, "x"])
+    ylim <- (c(0, max(yval, na.rm = TRUE)))
+    xlim <- c(min(c(0, xval)), max(xval))
+  } else if (x$method == "P-FPV") {
+    ylab <- "Sample size"
```
vandenman, line 526:
```suggestion
    ylab <- gettext("Sample size")
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    yval <- (x$model$data[, "x"])
+    ylim <- rev(c(0, max(yval, na.rm = TRUE)))
+    xlim <- c(min(c(0, xval)), max(xval))
+    step <- ((max(yval) - min(yval))/5)
+    yax  <- c(plyr::round_any(1/min(yval), 10^(sapply(round(1/min(yval)), nchar) - 1)),
+              plyr::round_any(1/seq(step, 4 * step, by = step), 10), plyr::round_any(1/max(yval), 10))
+  } else if (x$method == "D-FIV") {
+    ylab <- "Total events"
```
vandenman, line 534:
```suggestion
    ylab <- gettext("Total events")
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    yval <- (x$model$data[, "x"])
+    ylim <- rev(c(0, max(yval, na.rm = TRUE)))
+    xlim <- c(min(c(0, xval)), max(xval))
+    step <- ((max(yval) - min(yval))/4)
+    yax  <- c(plyr::round_any(1/min(yval), 10^(sapply(round(1/min(yval)), nchar) - 1)),
+              plyr::round_any(1/seq(step, 4 * step, by = step), 10), plyr::round_any(1/max(yval), 10))
+  } else if (x$method == "D-FAV") {
+    ylab <- "Total events"
```
vandenman, line 542:
```suggestion
    ylab <- gettext("Total events")
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    "D-FIV" = "funnelAsymmetryTestDebrayFIV"
+  )
+}
+.metamiscFitFunnelAsymmetryggPlot  <- function(x, ref, xlab = "Effect size",
+                                               confint = TRUE, confint.level = 0.1, confint.alpha = .50, confint.col = "skyblue") {
+
+  if (!inherits(x, "fat"))
+    stop("Argument 'x' must be an object of class \"fat\".")
+  if (confint.level < 0 | confint.level > 1) {
+    stop("Argument 'confint.level' must be between 0 and 1.")
```
vandenman, line 511:
```suggestion
    stop("Argument 'x' must be an object of class \"fat\".", domain = NA)
  if (confint.level < 0 | confint.level > 1) {
    stop("Argument 'confint.level' must be between 0 and 1.", domain = NA)
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+    shortcut,
+    "E-UW"  = "funnelAsymmetryTestEggerUW",
+    "E-FIV" = "funnelAsymmetryTestEggerFIV",
+    "M-FIV" = "funnelAsymmetryTestMacaskillFIV",
+    "M-FPV" = "funnelAsymmetryTestMacaskillFPV",
+    "P-FPV" = "funnelAsymmetryTestPeters",
+    "D-FIV" = "funnelAsymmetryTestDebrayFIV"
+  )
+}
+.metamiscFitFunnelAsymmetryggPlot  <- function(x, ref, xlab = "Effect size",
```
vandenman, line 505:
```suggestion
.metamiscFitFunnelAsymmetryggPlot  <- function(x, ref, xlab = gettext("Effect size"),
```

R/classicalpredictionperformance.R
```diff
@@ -0,0 +1,654 @@
...
+.metamiscFitFunnelAsymmetryNames <- function(shortcut) {
+  switch(
+    shortcut,
+    "E-UW"  = "Egger (unweighted)",
+    "E-FIV" = "Egger (multiplicative overdispersion)",
+    "M-FIV" = "Macaskill",
+    "M-FPV" = "Macaskill (pooled)",
+    "P-FPV" = "Peters",
+    "D-FIV" = "Debray"
+  )
```
vandenman, line 492: Should these be translated? If yes, I would not use them for indexing in jaspResults.

vandenman (changes requested):
I forgot about some translation stuff, see below.

FBartos:
ok. This should address the remaining issues.

vandenman:
Okay, one final thingy:

- [ ] Could you increase the default width of the forest plot? Just so the x-axis tick labels do not overlap by default.
alt text


Two other things that we may want to discuss and open an issue for so they can be solved at some later point in time:

1. Why those extra columns? I've never seen them get filled by any analysis. I can export the O:E ratios or C-statistics, but those end up in different columns. I'm not saying you should remove them (apparently you can't) but what is their purpose?

2. Sometimes, JAGS crashes like this:
image
In this case, I selected `O:E ratio` where I should have selected `C-statistic`. The error message is rather cryptic though. These kinds of errors happen quite frequently with e.g., the debug data set and aren't informative at all. Is it possible to detect when this happens and show a more informative error message instead?

FBartos:
1) yes, that's not great indeed, done

2.1) If you check the ``Export ...`` it will fill them with the model-based OE / c-stat estimates. The author wanted to have the functionality to extract them from the analysis.
I will probably have a master student implementing effect sizes calculations/transformations for the meta-analysis module, so we might want to move the functionality there. I will also discuss with Joris whether there might be another way of exporting columns from R to JASP, but this was the thing he suggested previously.

2.2) okay, I created an issue (https://github.com/jasp-stats/INTERNAL-jasp/issues/1569) for it and will look further into solving it:

vandenman:
Regarding 2.1, I still wonder why this is the only analysis that does it this way. Other analyses that export columns don't create new columns by default and only add them after the user enters a name for the columns. Did you compare your code to other analyses, e.g., machine learning or network analysis?

Also, if you first do one analysis, and then another, the second analysis does not fill the columns:
image

This happens because the exported column names should be unique so you probably can't fix this, but just so you know (there is an open issue about this).

FBartos:
There might have been a misunderstanding between me and Joris. That's a behavior that I tried to explain to him that I want and I have been advised to use this qml element which results in this behavior (possibly because the old one had some undesired properties). 

btw, just tried the Machine learning one but is bugged (https://github.com/jasp-stats/INTERNAL-jasp/issues/1570)

vandenman:
now that the ML issue is resolved, is the behavior in the ML module closer to what you want to implement?

FBartos:
yes, that's exactly the behavior I wanted to have, and it also uses the ``ComputedColumnField`` which really confuses me rn

vandenman:
Do you have time to look at this today? The rest of the code is fine and can be merged in my opinion, but this would be nice to fix.

FBartos:
yes, just working on it and trying to figure out why it works in ML and didn't work for me when I tried originally

FBartos:
ok, so I figured it out - the issue was that I set ``value`` directly instead of ``placeholderText``

vandenman (approved):

vandenman:
Feel free to merge this once the tests pass.

PR #77, opened 2021-09-30
Add seed option to Bayesian meta-analysis

sophieberkhout:
Requested by @FBartos as he got questions about this.

The `meta_bma` function apparently requires to use the `set.seed` function instead of a seed argument, as it passes the seed to rstan internally.


sophieberkhout:
Would it now also be possible to not skip the unit tests anymore?

inst/qml/qml_components/BayesianMetaAnalysisAdvanced.qml
```diff
@@ -130,6 +130,11 @@ Section
 			}
 		}
 
+		Group
+		{
+			SetSeed { }
+		}
+
```
FBartos, line 137: I think that it doesn't need to be in a group of its own.

R/bayesianmetaanalysiscommon.R
```diff
@@ -19,6 +19,8 @@
 
 .BayesianMetaAnalysisCommon <- function(jaspResults, dataset, ready, options) {
   
+  .setSeedJASP(options)
```
FBartos, line 22: it might be better to put it right before the fitting function, in case that the order of flow changes / there are other rng elements

FBartos (changes requested):

FBartos:
Yes, I think that having the unit tests, at least for one OS would be great

FBartos:
@vandenman , do you think that this could be considered a bug fix and we could merge it in if we restrict the OS-specific unit tests to windows?

sophieberkhout:
I added an if statement now so the unit tests only run if the system is Windows. Let's see if it works :)

FBartos:
if so, maybe changing the if statement to ``testthat::skip_on_os(c("mac", "linux", "solaris"))`` might be preferable

sophieberkhout:
Okay they pass so I'll change it `skip_on_os`

FBartos (approved):
I'm not sure whether we can still merge this in though.

vandenman:
hmmm, there's not really an issue this fixes right? It's also not really an ideal fix, because this
```r
testthat::skip_on_os(c("mac", "linux", "solaris"))
```
means that this analysis should only be touched by people working on Windows (as people on macOS and Linux can never update the unit tests).

I'd be in favor of postponing this, assuming we can find a better solution.

FBartos:
The issue was a lack of reproducibility when performing Bayesian meta-analysis.

> means that this analysis should only be touched by people working on Windows (as people on macOS and Linux can never update the unit tests).

I guess that till Stan fixes the samplers rng across OSes our hands are tied. Isn't this still an upgrade for the unit tests since there were no unit tests for the numerical results before (i.e., they were skipped on all OSes)? 

Maybe an alternative could be adding dispatching of the unit tests for multiple OSes, but that would make their updating even more painful.

It would be a nice thing to have since several people asked for it on a workshop but we can postpone it.

vandenman:
> I guess that till Stan fixes the samplers rng across OSes our hands are tied.

From https://mc-stan.org/docs/2_27/reference-manual/reproducibility-chapter.html

image

So I don't think they will ever fix this (it's not even seen as an issue in their eyes).

> Isn't this still an upgrade for the unit tests since there were no unit tests for the numerical results before (i.e., they were skipped on all OSes)?

After the first time someone else works on this code, I think that these tests will just be skipped again.

> It would be a nice thing to have since several people asked for it on a workshop but we can postpone it.

I agree it is an upgrade, but I'm just not too fond of the unit tests. We could also just merge this and then see how long it takes before somebody feels the need to skip the tests. If we go that way, we might want to leave a comment in the unit test code to skip the tests if they start failing.

FBartos:
hmm, what do you suggest as the best way forward then? 

Adding a comment to the unit tests and merging it in?

vandenman:
> Adding a comment to the unit tests and merging it in?

At least a comment, so that the next person knows they can ignore the tests if they start failing.

vandenman (approved):
Feel free to merge this once the tests pass!

FBartos:
Thanks Don!

PR #93, opened 2021-11-02
Removed the Unnecessary Section

amirmasoudabdol:
So, this should solve your problem with the Data section. Please test it beforehand, and make sure that the `RadioButtonGroup` title, etc is right. I just moved the controls, and remove the `Section`.


FBartos (approved):
Looks good to me, thanks!

One think that we should think about is the fact that I name the radiobutton "Input type" in RoBMA since one of the option is loading an already fitted object from R

FBartos:
Should this be merged in?

PR #98, opened 2021-11-09
remove dependency on outdated GitHub BayesTools

FBartos:
The new version of RoBMA crashes the analysis due to version mismatch. Installing BayesTools from CRAN should resolve the issue.

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,5 +30,4 @@ Imports:
   ggmcmc
 Remotes:
   jasp-stats/jaspBase,
-  jasp-stats/jaspGraphs,
-  FBartos/BayesTools@3095a6d0f80348b8e274cb2e03f31c3d5f118f0b
+  jasp-stats/jaspGraphs


AlexanderLyNL:
Does this crash occur on 0.16?

FBartos:
Yes, found this out while testing the .msi

AlexanderLyNL:
I've got a different error on mac
Screenshot 2021-11-09 at 10 52 48

FBartos:
that's the same error

amirmasoudabdol:
Is it fair to partially blame this on our lousy DESCRIPTION files where we don't really specify the version number of our dependencies, not even R? If that's in place, then the build will not fail before we create the JASP app.

FBartos:
Well, I apparently forgot to update hard set version in the description when I updated both packages.

amirmasoudabdol:
That's fine, and it happens but I'm talking about the `Imports` list on `DESCRIPTION`. I think since we are very particular with the version that we are shipping we should probably be more careful and explicit with the version of packages that we are using on the `DESCRIPTION` file.

JorisGoosen:
> That's fine, and it happens but I'm talking about the `Imports` list on `DESCRIPTION`. I think since we are very particular with the version that we are shipping we should probably be more careful and explicit with the version of packages that we are using on the `DESCRIPTION` file.

But yeah, R is really bad at actually adhering to these versionnumbers so it won't help too much.

JorisGoosen (approved):

PR #99, opened 2021-11-12
N Variable List should not be disabled when a variable is set

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1627
This just solves this issue. But the RoBMA still fails, not sure if it is a problem:
Screenshot 2021-11-12 at 14 10 04


boutinb:
My fix is OK, so I merge it, but @FBartos you should have a look if the analysis error is normal.

boutinb:
And we should understand why the unit test on Windows fails...

PR #115, opened 2022-01-28
fix influence measures error

FBartos:
fixes: https://github.com/jasp-stats/jasp-test-release/issues/1727
fixes: https://github.com/jasp-stats/jasp-test-release/issues/1722

The metafor namespace was not loaded properly when the fit object was already fitted and only accessed from JASP state. Calling the influence function from metafor package directly solves the issue.


vandenman (approved):
> The metafor namespace was not loaded properly when the fit object was already fitted and only accessed from JASP state. Calling the influence function from metafor package directly solves the issue.

Yeah, we have had similar issues with the BAS package where loading an object from the state does not load the namespace so R cannot find the S3 method...

PR #131, opened 2022-06-14
fixes metafor package updates

FBartos:
e.g.: https://github.com/wviechtb/metafor/issues/61
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1878

there are some `.v()` and `.unv()` in the code, it might be worth removing them in the future if they are indeed deprecated?

### DIFF
--- a/R/bayesianmetaanalysis.R
+++ b/R/bayesianmetaanalysis.R
@@ -18,21 +18,21 @@
 # Main function ----
 
 BayesianMetaAnalysis <- function(jaspResults, dataset, options) {
-  
+
   options[["module"]] <- "metaAnalysis"
-  
+
   # Ready: variables needed for the analysis (confidence interval missing)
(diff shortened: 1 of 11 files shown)


vandenman (changes requested):

vandenman:
the changes look fine but the unit tests fail for the robustbayesianmetaanalysis with error message: `Error: The 'model:priors' argument cannot be NULL.`. Any idea what that is about?

FBartos:
So, this should fix the Classical Meta-Analysis, but there still seem to be some broken unit tests in RoBMA.

On MacOS, there is an issue with JAGS:
`fatal error: /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't open input file: /usr/local/libexec/jags-terminal (No such file or directory)`

There is also some additional problem in the windows/linux tests that I will try to fix.

PR #133, opened 2022-06-26
fix the export file

FBartos:
oh, so, this is all that needed to be changed?


JorisGoosen (approved):
Yes!

Always good to turn on "show extension" in your filemanager program ;)

Kucharssim (approved):
The mac unit test fails due to apparently jags not found - @vandenman is this something that you know of?

I think this is not an issue for the current PR though, so I will merge it

PR #137, opened 2022-08-29
option renaming and upgrades

sophieberkhout:
Classical and Bayesian meta-analysis ([#2019](https://github.com/jasp-stats/INTERNAL-jasp/issues/2019), [#2014](https://github.com/jasp-stats/INTERNAL-jasp/issues/2014)).

I did not rename options with `visible:		module == "cochrane"`, because I assume that would break things in the cochrane module. @FBartos not sure if other options being renamed is a problem for that module.


sophieberkhout:
Oops forgot to update tests.. one sec.

sophieberkhout:
So the remaining problem with the unit tests (only for macOS) seems to be with the robust meta-analysis which I made no changes to.

I think this is the problem:
> fatal error: /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't open input file: /usr/local/libexec/jags-terminal (No such file or directory)

@boutinb you mentioned in the slack something about not finding the jags package. Could this be related?

inst/Upgrades.qml
```diff
@@ -0,0 +1,147 @@
...
+
+Upgrades
+{
+	Upgrade
+	{
+		functionName:	"ClassicalMetaAnalysis"
+		fromVersion:	"0.15"
+		toVersion:		"0.16.4"
+
+        // ClassicalMetaAnalysis.qml
```
Kucharssim, line 12: could you replace spaces with tabs in the qml files please?

inst/Upgrades.qml
```diff
@@ -0,0 +1,147 @@
+import QtQuick		2.12
+import JASP.Module	1.0
+
+Upgrades
+{
+	Upgrade
+	{
+		functionName:	"ClassicalMetaAnalysis"
+		fromVersion:	"0.15"
```
Kucharssim, line 9: given we'll merge this after this release, you can change it to `fromVersion: "0.16.4"` and `toVersion: "0.17"`

inst/qml/qml_components/BayesianMetaAnalysisPriors.qml
```diff
@@ -85,7 +85,7 @@ Section
 				DoubleField
 				{
 					label: 			qsTr("std:")
-					name: 			"informativeNormalStd"
+					name: 			"normalStd"
```
Kucharssim, line 88:
```suggestion
					name: 			"normalSd"
```

Kucharssim (changes requested):
Looks quite good, just a couple of minor comments.

> I did not rename options with visible: module == "cochrane", because I assume that would break things in the cochrane module. @FBartos not sure if other options being renamed is a problem for that module.

That should not be a problem for that module as currently `jaspCochrane` does not source the QML files directly but relies on copies of the original files from here, so you should not be able to break anything. So it would make more sense to do the changes here and the renaming in the Cochrane module would be done just by @FBartos sourcing the files directly and copying the relevant sections for the Upgrader once this PR is merged..

Kucharssim:
Could you [rebase](https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/git-guide.md#rebasing-your-repository-github-documentation-rebasing) your branch and resolve the conflicts? Let me know if you have any help with that!

PR #155, opened 2022-12-24
Options renaming

FBartos:
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2015
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2016
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2017
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2018
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2020
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2021

to fix https://github.com/jasp-stats/INTERNAL-jasp/issues/2021, the `qml_components/RobustBayesianMetaAnalysisPriors.qml` and `qml_components/RobustBayesianMetaAnalysisWeightfunctions.qml`, which are quite complex due to the multiple dispatching (I have no idea how to consistently rename them according to the guide. Maybe @Kucharssim will know? I can then connect them back in R


Kucharssim:
> which are quite complex due to the multiple dispatching (I have no idea how to consistently rename them according to the guide

I am not sure what do you mean exactly, it seems to me both components can be checked for consistency with the guide quite easily. For example:

- use camelCase, so `two-sided` -> `twoSided`
- you do not need to prefix with `par` (I don't see what function it provides since it's already grouped together in the ComponentList)
- not sure but it seems worthwhile to group all "Custom Ensemble Models" with a common prefix, so instead using `effect`, `effectNull`, `heterogeneity`, we would have `modelsEffect`, `modelsEffectNull`, `modelsHeterogeneity`, etc.

Otherwise the components already look good, if you polish these things it should be just fine

FBartos:
The main issue I had was with names of the prior distribution parameters, the guide uses `normalPriorMean` as an example. I assume that I cut `Prior` as it is within a list. However, the `mu` is used for both normal and t-distribution. In the first case, it is mean, in the second it is location. Should I then just name it `mu` irrespective of the distribution-based dispatch?

Kucharssim:
> Should I then just name it mu irrespective of the distribution-based dispatch?

Yeah that sounds fine in this case

FBartos:
ok, this should be it then

FBartos:
fixes: https://github.com/jasp-stats/jasp-test-release/issues/2215

FBartos:
@Kucharssim seems like we kinda forgot on this?

Kucharssim (approved):
Okay, after a bit of fiddling I managed to polish the names and make the upgrader work. It should be fine now.

The tests are failing due to other reasons than the option names: https://jasp-stats.github.io/actions-dashboard/. I will merge this PR now so that we get this out of the way - @FBartos, @sophieberkhout, please fix the tests, thanks!

PR #168, opened 2023-04-17
fix coef display

FBartos:
the original code -- somewhat convoluted for whatever reason -- did not show estimated coefficients for models with 1 continuous covariate and no intercept


FBartos:
the fails are not related to the fix:
linux fails on stan compilation for predictive performance
maxOS fails on JAGS compilation for RoBMA

FBartos:
like building up and checking whether it creates the table correctly? Yes

PR #174, opened 2023-04-24
fixes non-functioning diagnostics plots

FBartos:
fixes: https://github.com/jasp-stats/jasp-test-release/issues/2315
(this was introduced in options renaming)

What does not make sense is why both the 
`Trim-fill analysis` and `Diagnostic plots` produce the very same output. I would expect the `Trim-fill analysis` to produce actual Trim-fill analysis. Could you fix that in a separate PR?

### DIFF
--- a/R/classicalmetaanalysiscommon.R
+++ b/R/classicalmetaanalysiscommon.R
@@ -511,7 +511,7 @@
 # Plots
 .metaAnalysisPlotsContainer <- function(container, options, ready)  {
   if(!ready) return()
-  if(!options$forestPlot && !options$funnelPlot && !options$plotResidualseffectSize &&
+  if(!options$forestPlot && !options$funnelPlot && !options$diagnosticPlot &&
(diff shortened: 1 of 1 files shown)


sophieberkhout (approved):

sophieberkhout:
> What does not make sense is why both the
Trim-fill analysis and Diagnostic plots produce the very same output. I would expect the Trim-fill analysis to produce actual Trim-fill analysis. Could you fix that in a separate PR?

The difference is that the plots of the trim-fill analysis include the estimated missing studies. They are different for me. If they are the same for you, could you make an issue?

FBartos:
ah, I see. The current implementation, with the full matrix, is not really usable though as the figures are too small and I can't stretch them out even on 4K screen to make anything visible with 50 studies. We should probably separate the figures 1 by one

FBartos:
I will make a separate issue for that

PR #218, opened 2024-01-31
Add Binomial-Normal BMA Meta-Analysis

FBartos:
also now:
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2129
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1201

### DIFF
--- a/R/bayesianbinomialmetaanalysis.R
+++ b/R/bayesianbinomialmetaanalysis.R
@@ -90,7 +90,6 @@ BayesianBinomialMetaAnalysis <- function(jaspResults, dataset, options, state = 
   "advancedRemoveFailedModels", "advancedRemoveFailedModelsRHat",  "advancedRemoveFailedModelsRHatTarget", "advancedRemoveFailedModelsEss", "advancedRemoveFailedModelsEssTarget", "advancedRemoveFailedModelsMcmcError", "advancedRemoveFailedModelsMcmcErrorTarget", "advancedRemoveFailedModelsMcmcErrorSd", "advancedRemoveFailedModelsMcmcErrorSdTarget",
   "advancedRebalanceComponentProbabilityOnModelFailure", "seed", "setSeed"
 )
-
 # helper functions
 .bibmaCheckReady          <- function(options) {
   return(options[["successesGroup1"]] != "" && options[["successesGroup2"]] != "" && options[["observationsGroup1"]] != "" && options[["observationsGroup2"]] != "")
@@ -112,7 +111,7 @@ BayesianBinomialMetaAnalysis <- function(jaspResults, dataset, options, state = 
   # We don't wanna delete the bibma modele every time settings is change since bibma takes a lot of time to fit.
   # Therefore, we don't create dependencies on the fitted model (in cases when the model can be updated), but
   # on a notifier that tells us when there was the change. If possible, we don't refit the whole model,
-  # just update the neccessary parts.
+  # just update the necessary parts.
 
   if (is.null(jaspResults[["modelNotifier"]])) {
     modelNotifier <- createJaspState()
@@ -121,7 +120,6 @@ BayesianBinomialMetaAnalysis <- function(jaspResults, dataset, options, state = 
   }
 
   return()
-
 }
 .bibmaCleanModel          <- function(jaspResults) {
 
@@ -382,7 +380,14 @@ BayesianBinomialMetaAnalysis <- function(jaspResults, dataset, options, state = 
 
   if (is.null(jaspResults[["model"]])) {
     model <- createJaspState()
-    model$dependOn(.bibmaDependencies[.bibmaDependencies != "studyLabel"])
+    model$dependOn(.bibmaDependencies[!.bibmaDependencies %in% c(
+      # the excluded dependencies can be handled via the update function
+      "studyLabel",
+      "advancedRemoveFailedModels", "advancedRemoveFailedModelsRHat",  "advancedRemoveFailedModelsRHatTarget", "advancedRemoveFailedModelsEss", "advancedRemoveFailedModelsEssTarget",
+      "advancedRemoveFailedModelsMcmcError", "advancedRemoveFailedModelsMcmcErrorTarget", "advancedRemoveFailedModelsMcmcErrorSd", "advancedRemoveFailedModelsMcmcErrorSdTarget",
+      "advancedRebalanceComponentProbabilityOnModelFailure"
+    )]
+    )
     jaspResults[["model"]] <- model
     fit                    <- NULL
   } else {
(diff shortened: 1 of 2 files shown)


R/bayesianmetaanalysisbinomial.R
```diff
@@ -0,0 +1,457 @@
...
+  if (options[["priorDistributionPlot"]])
+    .bibmaPriorsPlots(jaspResults, options)
+
+
+  ### Inference
+  # default summary
+  .robmaSummaryTable(jaspResults, options, type = "BiBMA")
+  # models overview
+  if (options[["inferenceModelsOverview"]])
+    .robmaModelsOvervievTable(jaspResults, options, type = "BiBMA")
```
juliuspfadt, line 50:
```suggestion
    .robmaModelsOverviewTable(jaspResults, options, type = "BiBMA")
```

R/bayesianmetaanalysisbinomial.R
```diff
@@ -0,0 +1,457 @@
...
+  # extract the priors
+  if (is.null(jaspResults[["model"]]))
+    priors <- jaspResults[["priors"]][["object"]]
+  else
+    priors <- jaspResults[["model"]][["object"]][["priors"]]
+
+  # create container for each of the parameters
+  for (parameter in c("effect", "heterogeneity")) {
+
+    if (!is.null(priorPlots[[parameter]]))
```
juliuspfadt, line 220: I would prefer that if you work with if else and one of them needs brackets that the other also gets brackets.

juliuspfadt:
I am not at all done, some preliminary things maybe: 

- I get an empty plot with this file. I have no idea if what I was doing made sense, the only data set I could find was the metaAnalsyis stuff form the data library
[emptyplot.jasp.zip](https://github.com/jasp-stats/jaspMetaAnalysis/files/14195847/emptyplot.jasp.zip)
- I do not like the name of the analysis in JASP. Why not call it "Binomial Meta-Analysis"? The "Binomial" in brackets just makes me think what should be in brackets for the Meta Analysis above it. But maybe that is the actual issue. If one if Binomial what is the other?

inst/qml/qml_components/RobustBayesianMetaAnalysisAdvanced.qml
```diff
@@ -0,0 +1,244 @@
...
+			}
+
+			CheckBox
+			{
+				label: 				qsTr("Effective sample size")
+				name:				"advancedAutofitEss"
+				checked:			true
+				childrenOnSameRow:	true
+
+				DoubleField
```
juliuspfadt, line 137: I wonder if it makes sense to have this depend on the number of samples?

inst/qml/qml_components/RobustBayesianMetaAnalysisAdvanced.qml
```diff
@@ -0,0 +1,244 @@
...
+				{
+					name:			"advancedAutofitEssTarget"
+					defaultValue:	500
+					min:			1
+					inclusive:		JASP.None
+				}
```
juliuspfadt, line 143: I wonder if it makes sense to have this depend on the number of samples?

inst/qml/qml_components/RobustBayesianMetaAnalysisAdvanced.qml
```diff
@@ -0,0 +1,244 @@
...
+						DropDown
+						{
+							name:			"advancedAutofitMaximumFittingTimeTargetUnit"
+							values:
+							[
+								{ label: qsTr("hours"),				value: "hours"},
+								{ label: qsTr("minutes"),			value: "mins"},
+								{ label: qsTr("seconds"),			value: "secs"}
+							]
+						}
```
juliuspfadt, line 205: I am curious, how do you measure the time?

inst/qml/qml_components/RobustBayesianMetaAnalysisDiagnostics.qml
```diff
@@ -0,0 +1,114 @@
...
+			name:		"mcmcDiagnosticsPlotPeese"
+			visible:	analysisType === "RoBMA"
+		}
+	}
+
+	Group
+	{
+		Group
+		{
+			title:			qsTr("Type")
```
juliuspfadt, line 80: So are all these plots? Is it the plot type of the Effect and Heterogeneity plots? If that is the case it should be better visible.

juliuspfadt (changes requested):
Very nice! 
See the comments for potential improvements

R/bayesianmetaanalysisbinomial.R
FBartos: lol thanks!

R/bayesianmetaanalysisbinomial.R
```diff
@@ -0,0 +1,457 @@
...
+  } else {
+    # only tries to update labels if they weren't supplied originally
+    fit <- update(fit, study_names = if (options[["studyLabel"]] != "") dataset[, options[["studyLabel"]]])
+
+  }
+
+
+  # error handling
+  if (jaspBase::isTryError(fit))
+    .quitAnalysis(fit)
```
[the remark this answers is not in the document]
FBartos: not really--this should never crash, but if it does, it at least gives users some package generated message to return

inst/qml/qml_components/RobustBayesianMetaAnalysisAdvanced.qml
FBartos: yeah, that would make sense, but I believe we don't do it at other places and I wouldn't like to have the autofit options change automatically once I adjust the rest of MCMC settings

inst/qml/qml_components/RobustBayesianMetaAnalysisAdvanced.qml
FBartos: `runjags` R package does that

inst/qml/qml_components/RobustBayesianMetaAnalysisDiagnostics.qml
FBartos: I'm not sure what you mean actually

FBartos:
I fixed everything but one unit test--it works well when running interactively but running it via testAnalysis always creates slightly different results. No idea what's the issue though (I updated all packages etc).

FBartos:
the convergence check function does not refit the model, but the `update()` function in RoBMA does not update the convergence assessment properly (https://github.com/FBartos/RoBMA/issues/34)

I would just merge the PR and I will fix the package side bug separately.

juliuspfadt (approved):
sorry for taking so long

PR #252, opened 2024-11-09
fix i18n: use gettextf and `%` together

shun2wang:


R/effectsizecomputation.R
```diff
@@ -51,7 +51,7 @@ EffectSizeComputation <- function(jaspResults, dataset, options, state = NULL) {
 
     # set error message if reported effect sizes cannot be performed
     if (effectSizeType[["design"]] == "reportedEffectSizes" && !.escReportedEffectSizesReady(variables, all = TRUE)) {
-      newDataOutput <- try(stop(gettext("Cannot compute outcomes. Chech that all of the required information is specified via the appropriate arguments (i.e. an Effect Size and either Standard Error, Sampling Variance, or 95% Confidence Interval).")))
```
FBartos, line 54: Are you sure that here and below should be gettextf()? The `%` sign here refers to the percentage symbol rather than string etc dispatch? The changes in clasicalmetaanalysiscommon and clasicalmetaanalysismultilevelmultivariate are good

shun2wang: Yes, really sure. see 3b in:https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-translation-rules.md
this rule also included in i18n check.

FBartos: oh, I didn't know, thanks!

FBartos (approved):

PR #303, opened 2025-04-23
small fixes

FBartos:
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2734
- better handling of cluster size reporting
- better reporting of very large test statistics
- added error messages for clustering non-convergence

### DIFF
--- a/R/classicalmetaanalysiscommon.R
+++ b/R/classicalmetaanalysiscommon.R
@@ -3986,7 +3986,13 @@
     heterogeneityName <- gettextf("Heterogeneity")
   }
 
-  return(sprintf("%1$s: Q(%2$i) = %3$.2f, %4$s", heterogeneityName, fit[["k"]] - fit[["p"]], fit[["QE"]], .maPrintPValue(fit[["QEp"]])))
+  return(sprintf(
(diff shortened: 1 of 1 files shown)


JohnnyDoorn:
I'll review this when the dev modules are fixed because at the moment i cannot load it in any version of jasp i have..

JohnnyDoorn:
I just refreshed the .jasp file from my issue, and some has been fixed, but I'm still getting an error for the first analysis:
image

I also found the message for "modifying the clustering settings" to be a bit vague - which options are you referring to? One of clubsandwich or small sample correction? Because I couldn't get any option to work (which I guess can be fine, but then the message could maybe be updated)
image

FBartos:
thanks Johnny!

I have a bigger PR with additional fixes coming, so I will fix the new error message there (or I will die of conflicts).
You are right that the error message is a bit vague but that's probably as much guidance I can do: either one of these changes helps or you need to remove clustering all together :)

Cheers,
Frantisek

JohnnyDoorn:
OK then I guess this is good to go! I asked about the clustering setting, because I changed those options but the error persisted, but I guess sometimes it cannot be helped.

PR #306, opened 2025-04-30
add variance-covariance matrix computation

FBartos:


FBartos:
it's a bit difficult to specify the analyses correctly, so here are two examples: 
[knapp2017.zip](https://github.com/user-attachments/files/19974087/knapp2017.zip)
[berkey1998.zip](https://github.com/user-attachments/files/19974088/berkey1998.zip)
(unzip)

koenderks (approved):

PR #321, opened 2025-06-06
Release Fixes

FBartos:
- fixes funnel plot error handling


FBartos:
which is now required to dispatch the analysis method estimator

PR #340, opened 2025-07-10
add Mantel-Haenszel / Peto

FBartos:
adds Mantel-Haenszel / Peto meta-analytic methods
(with beta tag)


juliuspfadt:
[error1.jasp.zip](https://github.com/user-attachments/files/21179139/error1.jasp.zip)

juliuspfadt (changes requested):
see previous messages :)

juliuspfadt:
that's not how it should be done....

PR #406, opened 2026-02-09
Fix Footnote Error

FBartos:
The following fix was achieved by:

> Examine the JASP file in C:\Users\fbart\Downloads\test_meta_3. It results in an error. Replicate the error, identify the issue, and fix the issue.

- Claude Code / Copilot spins its own R session
- reproduces the bug
- implements the fix
- verifies that the analysis runs

fixes: https://github.com/jasp-stats/jasp-test-release/issues/3134


vandenman (approved):
The changes are good, but why are there two .mcp.json files? In general, I think that adding LLM infrastructure should be done in a separate PR from fixes. Since this is not a common module I guess it's fine this time.

PR #436, opened 2026-06-03
Add GLMM, Risk of Bias, Effect Size Aggregation, Forest Plot, and other fixes

### LINKED ISSUE #3581: [Bug]: Meta-Analysis: Forest Plot Removes Aesthetics Reaching Outside Of Set Plotting Range
### JASP Version

0.95.0

### Commit ID

_No response_

### JASP Module

Meta Analysis

### What analysis are you seeing the problem on?

Meta-Analysis: All

### What OS are you seeing the problem on?

Windows 11

### Bug Description

Setting shorter limit than some of the CIs removes said CIs

### Expected Behaviour

Keep them aethetics outside of plotting range in

### Steps to Reproduce

Forest plot - > reduce plotting range

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

- [x] I have included a screenshot showcasing the issue, if possible.
- [x] I have included a JASP file (zipped) or data file that causes the crash/bug, if applicable.
- [x] I have accurately described the bug, and steps to reproduce it.

FBartos:
## Added
* Added Classical Generalized (GLMM) Meta-Analysis.
* Added Effect Size Aggregation for clustered/dependent effect sizes.
* Added Forest Plot analysis.
* Added Risk of Bias Plot analysis for RoB 2, ROBINS-I, and QUADAS-2.

## Changed
* Improved forest plots with transformed axis labels, manual ticks, axis limits, reference lines, weights, metadata panels, and subgroup/aggregation displays.
* Improved forest plot handling of estimates and intervals outside the plotting range.
* Updated dependencies and test coverage for new analyses and plots.

## Fixed
* Fixed footnote handling in classical meta-analysis moderator-test tables.
* Improved/fixed transformation notes for nonlinear estimates and marginal means.
* Fixed small help/output typos.


juliuspfadt (changes requested):
Resolve conflicts.

juliuspfadt:
builder bot build for glory

RensDofferhoff:
build

RensDofferhoff:
build

juliuspfadt (commented):
## Code Review

Reviewed the diff manually across the key new files. See inline comments below for specific findings. Overall the new analyses are well-structured and follow project conventions.

Main concerns: a potential sort-inversion in forest plot ordering, a crash path in the GLMM Wald test, and silent-failure edge cases in Effect Size Aggregation.

Files not fully reached due to diff size: `R/classicalmetaanalysiscommon.R` (1002+/174−), `R/forestplotcommon.R` (2529 lines), `R/effectsizecomputation.R` (402 lines).

R/forestplotdata.R
```diff
@@ -0,0 +1,1126 @@
...
+
+  if (options[["forestPlotStudyInformationOrderBy"]] == "") {
+    return(dfForest)
+  }
+
+  # "Ascending" means values increase top-to-bottom (forest plot convention),
+  # which corresponds to decreasing = TRUE in R's order() (row 1 = top of plot).
+  dfForest <- dfForest[order(
+    dfForest[[options[["forestPlotStudyInformationOrderBy"]]]],
+    decreasing = options[["forestPlotStudyInformationOrderAscending"]]
```
juliuspfadt, line 147: **Potential sort-direction inversion**

The comment above says _'Ascending means values increase top-to-bottom, which corresponds to `decreasing = TRUE` in R's `order()`'_ — but `decreasing = TRUE` puts the **largest** value in row 1. If row 1 maps to the top of the plot, values will **decrease** top-to-bottom, which is the opposite of ascending.

When the user ticks "Ascending" (option = `TRUE`), this line passes `decreasing = TRUE` and sorts the largest value to the top. Please add a snapshot test that explicitly verifies ordering direction, or flip the assignment to `decreasing = !options[["forestPlotStudyInformationOrderAscending"]]` if the current behavior is inverted.

R/classicalgeneralizedmetaanalysis.R
```diff
@@ -0,0 +1,469 @@
...
+# predicates
+.maIsGLMM <- function(options) {
+  options[["analysis"]] == "generalizedMetaAnalysis"
+}
+
+# anova.rma is not available for rma.glmm; compute Wald-type tests manually
+.maGlmmWaldTest <- function(fit, btt) {
+  beta <- as.vector(fit$beta[btt])
+  vb   <- fit$vb[btt, btt, drop = FALSE]
+  QM   <- as.vector(t(beta) %*% chol2inv(chol(vb)) %*% beta)
```
juliuspfadt, line 43: **Unguarded matrix inversion — cryptic crash on near-singular `vb`**

`chol()` throws a low-level LAPACK error `"matrix is not positive definite"` when `vb` is near-singular, e.g. with a saturated model (all df consumed by moderators) or very few studies. The user gets no actionable message.

Suggest:
```r
QM <- tryCatch(
  as.vector(t(beta) %*% chol2inv(chol(vb)) %*% beta),
  error = function(e) .quitAnalysis(gettext(
    "Wald test failed: variance matrix is singular. This can occur with saturated models or too few studies."
  ))
)
```

R/effectsizeaggregation.R
```diff
@@ -0,0 +1,429 @@
...
+
+  esaState <- createJaspState()
+  esaState$dependOn(.esaDependencies)
+  jaspResults[["esaState"]] <- esaState
+
+  if (!ready)
+    return()
+
+  # build sampling variance from the SE column
+  vi <- dataset[[options[["effectSizeStandardError"]]]]^2
```
juliuspfadt, line 137: **Missing SE > 0 validation**

A column with zero or negative SE values will silently produce zero or positive `vi` (squaring removes the sign). Downstream `aggregate.escalc` may then produce wrong results or a cryptic error with no clear source.

Add before this line:
```r
se <- dataset[[options[["effectSizeStandardError"]]]]
if (any(se <= 0, na.rm = TRUE))
  .quitAnalysis(gettext("All standard errors must be positive."))
vi <- se^2
```

R/effectsizeaggregation.R
```diff
@@ -0,0 +1,429 @@
...
+  clusterVar    <- options[["cluster"]]
+  nRows         <- nrow(dataset)
+
+  # map aggregated values back to first row of each cluster
+  yiOut <- rep(NA_real_, nRows)
+  viOut <- rep(NA_real_, nRows)
+
+  aggCluster <- aggResult[[clusterVar]]
+  for (i in seq_len(nrow(aggResult))) {
+    firstRow       <- which(clusterValues == aggCluster[i])[1]
```
juliuspfadt: **Silent no-op when cluster value is NA**

```r
firstRow <- which(clusterValues == aggCluster[i])[1]
```

If `aggCluster[i]` is `NA` (e.g. the aggregated result has an NA cluster label), then `clusterValues == NA` is entirely `NA`, `which()` returns `integer(0)`, and `[1]` returns `NA`. In R, `yiOut[NA] <- value` is silently ignored — the aggregated effect size is lost from the output column with no warning.

Consider guarding with:
```r
firstRow <- which(clusterValues == aggCluster[i])[1]
if (!is.na(firstRow)) {
  yiOut[firstRow] <- as.numeric(aggResult$yi[i])
  viOut[firstRow] <- as.numeric(aggResult$vi[i])
}
```

R/classicalgeneralizedmetaanalysis.R
```diff
@@ -0,0 +1,469 @@
...
+  zval <- Xb / se
+  if (!is.null(fit$test) && fit$test == "t") {
+    ddf  <- fit$k - fit$p
+    pval <- 2 * pt(abs(zval), df = ddf, lower.tail = FALSE)
+  } else {
+    pval <- 2 * pnorm(abs(zval), lower.tail = FALSE)
+  }
+  if (adjust != "none")
+    pval <- p.adjust(pval, method = adjust)
+  list(zval = zval, pval = pval, ddf = if (!is.null(fit$test) && fit$test == "t") ddf else NULL)
```
juliuspfadt, line 71: **`ddf = NULL` in named list — silent instead of safe**

```r
list(zval = zval, pval = pval, ddf = if (!is.null(fit$test) && fit$test == "t") ddf else NULL)
```

Returning `NULL` for a named list field is unusual: callers that do `result$ddf` get `NULL` silently, which is indistinguishable from a missing field and can propagate invisibly. Use `NA_real_` so callers can check with `is.na()`:

```r
ddf = if (!is.null(fit$test) && fit$test == "t") ddf else NA_real_
```

R/riskofbiasplot.R
```diff
@@ -0,0 +1,272 @@
...
+    data <- data[is.na(data[["domain"]]) | data[["domain"]] != "Overall", , drop = FALSE]
+    if (is.factor(data[["domain"]]))
+      data[["domain"]] <- droplevels(data[["domain"]])
+    return(data)
+  }
+
+  if (is.data.frame(plotObj[["data"]]) && "domain" %in% names(plotObj[["data"]]))
+    plotObj[["data"]] <- dropOverall(plotObj[["data"]])
+
+  layers <- plotObj[["layers"]]
```
juliuspfadt, line 143: **Fragile access to ggplot2 internal `$layers`**

```r
layers <- plotObj[["layers"]]
```

This patches ggplot2's internal list structure directly. It works today but will break silently if robvis changes which data it stores in layers, or if ggplot2 changes the layer representation. There is no robvis version pin here to detect regressions.

At minimum, add a version-aware comment:
```r
# Workaround: robvis does not expose an overall=FALSE API.
# Tested against robvis 0.3.0. Re-check on robvis updates.
```
Ideally, file an upstream issue with robvis to add a proper `include_overall` argument.

R/effectsizeaggregation.R
```diff
@@ -0,0 +1,429 @@
...
+
+    # save V matrix if requested
+    savePath <- options[["varianceCovarianceMatrixSaveComputedVarianceCovarianceMatrix"]]
+    if (!is.null(savePath) && savePath != "") {
+      try(write.table(vMatrix, file = savePath, sep = ",", row.names = FALSE, col.names = FALSE))
+    }
+      
+  }
+
+  aggResult <- try(do.call(metafor::aggregate.escalc, aggArgs))
```
juliuspfadt, line 224: **Inconsistent `try()` vs `tryCatch`**

`tryCatch` is used at lines 113 and 123 in this same function, but `try()` is used here. Minor, but worth harmonizing so the error-handling pattern is consistent throughout the file.

R/forestplot.R
```diff
@@ -1,1714 +1,618 @@
...
+    forestPlotStudyInformationAggregateMethod                       = "boxplot",
+    forestPlotStudyInformationAggregateMethodBubbleRelativeSize     = 1,
+    forestPlotStudyInformationBoxplotWeightsPercentage              = TRUE,
+    forestPlotStudyInformationBoxplotWeightsNormalized              = TRUE,
+    # shared component: estimate information
+    forestPlotEstimateInformationSelectedVariables                  = list(),
+    forestPlotEstimateInformationSelectedVariablesSettings          = list(),
+    # shared component: settings
+    forestPlotEstimatesAndConfidenceIntervals                       = TRUE,
+    forestPlotAllignLeftPanel                                       = "right",
```
juliuspfadt: **Typo: `Allign` → `Align`**

```r
forestPlotAllignLeftPanel = "right",
```

The same typo appears in the QML `name:` field (`forestPlotAllignLeftPanel` in `inst/qml/qml_components/ForestPlotSettings.qml` line 62). Since R and QML must agree on the key string exactly, both need to be fixed in the same commit — or left as-is if there are already saved .jasp files that depend on this key name (in which case an `Upgrades.qml` entry is needed).

DESCRIPTION
```diff
@@ -31,6 +31,7 @@ Imports:
   metamisc (>= 0.2.5),
   ggmcmc,
   pema,
+  robvis,
```
juliuspfadt, line 35: **New `robvis` hard dependency — confirm CI availability**

`robvis` is now a required import. Please confirm it is available in the JASP CI build environment (i.e., it is on CRAN or added to `Remotes:`). If it is not on CRAN yet, the package build will fail for other contributors.

DESCRIPTION
```diff
@@ -31,6 +31,7 @@ Imports:
...
   clubSandwich,
   CompQuadForm,
   sp,
```
juliuspfadt, line 38: **`sp` is a legacy spatial package**

`sp` has been superseded by `sf` and is in maintenance mode. Please check whether any code in this PR actually calls `sp` functions — if not, it can be removed. If it is still required (e.g. as a transitive dependency of another package), add a comment explaining why.

juliuspfadt:
**`.claude/settings.local.json` — local developer settings should not be committed**

This file (note the `local` suffix — a Claude Code convention for machine-specific config) contains personal permission allowlists and MCP server names. Committing it means every contributor who uses Claude Code on this repo inherits your permissions.

Add `.claude/settings.local.json` to `.gitignore`. If any permissions should be shared across contributors, move them to `.claude/settings.json` instead.

juliuspfadt:
## Bugs in  (pre-existing, unchanged by this PR)

These lines were not modified by this PR so GitHub prevents inline comments, but the bugs affect effect-size types that are central to the module.

---

### 1. Typo in error message — line 57

```r
newDataOutput <- try(stop(gettextf("Cannot compute outcomes. Chech that all of the required information...")))
```

**"Chech"** should be **"Check"**.

---

### 2. Duplicate `ti` key silently breaks PCOR/ZPCOR computation — lines 404–412

```r
if (effectSize %in% c("PCOR", "ZPCOR")) {
  inputs <- list(
    ti = dataset[[variables[["tStatistic"]]]],
    mi = dataset[[variables[["predictors"]]]],
    ni = dataset[[variables[["sampleSize"]]]],
    ti = dataset[[variables[["tStatistic"]]]],  # ← duplicate key
    ri = dataset[[variables[["semipartialCorrelation"]]]],
    pi = dataset[[variables[["pValue"]]]]
  )
}
```

In R, a named list can hold duplicate keys, but when this list is passed via `do.call(metafor::escalc, inputs)` the call resolves to `escalc(..., ti = x, ti = x, ...)`, which produces **"Error: formal argument 'ti' matched by multiple actual arguments"**. The `try()` wrapper catches this silently, so PCOR/ZPCOR effect sizes always fail without any visible error.

Same bug on line 419 for SPCOR/ZSPCOR: `ti` appears twice (the second should not be there; `r2i` — which is the distinctive input for this branch — is already present one line earlier).

---

### 3. Typo `"ZHI"` instead of `"ZPHI"` — line 527

```r
(design == "variableAssociation" && measurement == "binary" && effectSize %in% c("PHI", "ZHI")) ||
```

`"ZHI"` is not a valid metafor effect size name; the correct name is `"ZPHI"`. Because the condition never matches, `vtype` is never added for ZPHI computations, causing metafor to use its default (possibly wrong) sampling-variance formula for that measure.

juliuspfadt:
it would be nice to have one or two datasets to test this with in the PR

juliuspfadt:
or rather even a .asp file, since there are no linked issues that I can check with. I was trying this with JASP-Ai but that failed miserably in even starting a meta analysis.

FBartos:
> or rather even a .asp file, since there are no linked issues that I can check with. I was trying this with JASP-Ai but that failed miserably in even starting a meta analysis.

there is a set of JASP files in the testthat folder under "other" :)

NEWS
```diff
@@ -0,0 +1,34 @@
...
+
+---
+
+# jaspMetaAnalysis 0.95.5
+
+## Added
+* Added Classical Generalized (GLMM) Meta-Analysis.
+* Added Effect Size Aggregation for clustered/dependent effect sizes.
+* Added Forest Plot analysis.
+* Added Risk of Bias Plot analysis for RoB 2, ROBINS-I, and QUADAS-2.
```
juliuspfadt, line 23: This reads kind of cryptic to me, can you specify which analyses this corresponds to?

juliuspfadt:
I found a somewhat uninformative error with one of the examples: 
[effectSizeError.jasp.zip](https://github.com/user-attachments/files/28945063/effectSizeError.jasp.zip)
Screenshot 2026-06-15 at 09 13 28

juliuspfadt:
About effect size aggregation, what exactly is the main output of the analysis? is it the cov matrix that can be saved? If so, maybe this should be more clear, either by having the top section always unfolded, or moving the save field somehow further up. 
or wait, it computes columns? maybe this is obvious to people using the analysis... but not to me, so maybe it helps to move the naming fields for the new variables into a more prominent place?

DESCRIPTION
FBartos: bad bot:
https://cran.r-project.org/web/packages/robvis/

juliuspfadt:
Another error that could be more informative: 
Screenshot 2026-06-15 at 09 31 09
[glmmError.jasp.zip](https://github.com/user-attachments/files/28945623/glmmError.jasp.zip)

R/riskofbiasplot.R
FBartos: not fix

R/classicalgeneralizedmetaanalysis.R
FBartos: not fix

R/forestplotdata.R
FBartos: there are tests for this, not fix

DESCRIPTION
FBartos: required for metafor mv functionality

juliuspfadt:
[glmmError2.jasp.zip](https://github.com/user-attachments/files/28945801/glmmError2.jasp.zip)
Screenshot 2026-06-15 at 09 37 31

juliuspfadt:
Screenshot 2026-06-15 at 09 38 09

[glmmError3.jasp.zip](https://github.com/user-attachments/files/28945845/glmmError3.jasp.zip)

juliuspfadt:
There should be column headings in the forest plot for the first two variables, that is, author and year, also if its year maybe the format of the column should not be a regular number. I used the example from library (see the filename at the top of the screenshot)
Screenshot 2026-06-15 at 09 50 01

juliuspfadt:
I refreshed the forest plot standalone analysis and it errored: 
Screenshot 2026-06-15 at 09 53 21
[forestError.jasp.zip](https://github.com/user-attachments/files/28946369/forestError.jasp.zip)

juliuspfadt:
I tried the risk of bias example and got some empty plots and an error: 
Screenshot 2026-06-15 at 09 56 28
Screenshot 2026-06-15 at 09 56 20
[riskError.jasp.zip](https://github.com/user-attachments/files/28946499/riskError.jasp.zip)

FBartos:
> I tried the risk of bias example and got some empty plots and an error:

this is desired behavior untill you specify at least one domain of the risk of bias -- otherwise there is no content that could be plotted

FBartos:
> There should be column headings in the forest plot for the first two variables, that is, author and year, 

not true, the title is a separate input field (empty) for those two variables
 
> also if its year maybe the format of the column should not be a regular number. I used the example from library (see the filename at the top of the screenshot)

yes, ideally, this is an issue how the data are forwarded from JASP, no idea how to solve those

FBartos:
> About effect size aggregation, what exactly is the main output of the analysis? is it the cov matrix that can be saved? If so, maybe this should be more clear, either by having the top section always unfolded, or moving the save field somehow further up.
or wait, it computes columns? maybe this is obvious to people using the analysis... but not to me, so maybe it helps to move the naming fields for the new variables into a more prominent place?

I extended the table note to make it clear that the columns were added to the dataset

NEWS
FBartos: stays as is

juliuspfadt:
build builder bot build for glory

juliuspfadt:
> I found a somewhat uninformative error with one of the examples: [effectSizeError.jasp.zip]

this is fixed, however I wonder if there should be some check on the size of the subgroups? This is not very important, just a small note

juliuspfadt:
this seems like its not fixed: https://github.com/jasp-stats/jaspMetaAnalysis/pull/436#issuecomment-4705738320
I see a similar error with this 
[forestErrorAgain.jasp.zip](https://github.com/user-attachments/files/29098263/forestErrorAgain.jasp.zip)
Screenshot 2026-06-18 at 16 20 58

and the old file also still errors

R/effectsizecomputation.R
```diff
@@ -843,6 +871,381 @@ EffectSizeComputation <- function(jaspResults, dataset, options, state = NULL) {
...
+          m2i  = "meanGroup2",
+          sd1i = "sdGroup1",
+          sd2i = "sdGroup2",
+          n1i  = "sampleSizeGroup1",
+          n2i  = "sampleSizeGroup2",
+          ti   = "tStatistic",
+          pi   = "pValue",
+          di   = "cohensD"
+        )
+      } else if (effectSize %in% c("SMD1", "SMDH1")) {
```
juliuspfadt: `"SMDH1"` is a transposition of `"SMD1H"` — the canonical dropdown value defined in `EffectSizeComputation.qml:144` is `"SMD1H"`. This branch is therefore dead code: an SMD1H effect size falls through to the `else` path, which passes `sd1i` to `escalc`. SMD1H only needs `sd2i` (single control SD), so passing `sd1i` causes incorrect effect size computation or an unexpected error that is caught silently by `try()`.

Same typo at lines 201 and 558 (unchanged context). All three should read `c("SMD1", "SMD1H")`.

`EffectSizeComputation.qml:638` also uses `"SMDH1"` in a visibility binding — that element is incorrectly visible/hidden when SMD1H is selected.

R/classicalmetaanalysiscommon.R
```diff
@@ -1165,9 +1228,11 @@ ClassicalMetaAnalysisCommon <- function(jaspResults, dataset, options, ...) {
   # .maGetEstimatedMarginalMeansAndContrastsOptions()
 
   # check whether the section should be created at all
-  isReadyEffectSize    <- (length(options[["estimatedMarginalMeansEffectSizeSelectedVariables"]])    > 0 || options[["estimatedMarginalMeansEffectSizeAddAdjustedEstimate"]])    &&
+  isReadyEffectSize    <- .maIsMetaregressionEffectSize(options) &&
```
juliuspfadt, line 1232: Adding `.maIsMetaregressionEffectSize(options)` as the first condition makes `isReadyEffectSize` always `FALSE` for plain meta-analyses (no moderators in the effect-size model), because `length(options[["effectSizeModelTerms"]]) == 0`. The entire EMM section is then silently removed/never created at lines 1238–1244.

In the old code (master) this was not a prerequisite — EMM output was shown for plain meta-analyses whenever the user enabled it. The QML does disable the Effect Size EMM controls when there are no moderators, but `estimatedMarginalMeansEffectSizeAddAdjustedEstimate` is separate, and more importantly old `.jasp` files with this option saved will now silently produce no output.

R/classicalmetaanalysiscommon.R
```diff
@@ -837,6 +882,15 @@ ClassicalMetaAnalysisCommon <- function(jaspResults, dataset, options, ...) {
...
   }
 
+  if (.maIsGLMM(options) && .maIsMetaregressionEffectSize(options)) {
+    if (options[["fixedEffectTest"]] == "t")
+      testsTable$addFootnote(gettext("Moderation test based on a Wald-type F-test."))
+    else
+      testsTable$addFootnote(gettext("Moderation test based on a Wald-type chi-squared test."))
+  }
+
+  .maAddLowDdfWarning(testsTable, fit, options)
```
juliuspfadt, line 892: `.maAddLowDdfWarning` is a new function added in this PR but it is only called once — here, for the hypothesis tests table. The meta-regression **coefficients** table (`.maCoefficientEstimatesTable`, ~line 1080) shows per-coefficient p-values under the same knha/club-sandwich inference, but never receives this low-df footnote. When effective df < 4, the coefficients table displays potentially unreliable p-values with no caution warning while the tests table correctly shows one.

R/classicalgeneralizedmetaanalysis.R
```diff
@@ -0,0 +1,475 @@
...
+    if (options[["heterogeneityTau"]])  1,
+    if (options[["heterogeneityTau2"]]) 2,
+    if (options[["heterogeneityI2"]])   3,
+    if (options[["heterogeneityH2"]])   4
+  )
+
+  heterogeneity <- heterogeneity[heterogeneityShow, , drop = FALSE]
+
+  if (!options[["confidenceIntervals"]])
+    heterogeneity <- heterogeneity[, c("par", "est")]
```
juliuspfadt: When `standardErrors = TRUE` and `confidenceIntervals = FALSE`, `heterogeneity$se` is added at line 306, but this subsetting to `c("par", "est")` unconditionally drops it — the SE column is never surfaced in the GLMM heterogeneity table regardless of what the user selected. The condition should be `c("par", "est", if (options[["standardErrors"]] && !is.na(fit[["se.tau2"]])) "se")`.

R/effectsizeaggregation.R
```diff
@@ -0,0 +1,637 @@
...
+    if (options[["varianceCovarianceMatrixSimpleStructure"]] == "CS*CAR" && length(constructVar) > 0 && !is.na(constructVar) && constructVar != "")
+      aggInput$obs <- .esaAsRName(constructVar)
+  } else if (tryCatch(.mammVarianceCovarianceMatrixReady(options), error = function(e) FALSE)) {
+    vcalcInput       <- .mammGetVarianceCovarianceMatrix(NULL, options, returnCall = TRUE)
+    vcalcInput$data  <- as.name("dat")
+
+    aggInput$V <- "effectSizeVarianceCovarianceMatrix"
+
+    if (options[["varianceCovarianceMatrixType"]] == "precomputed") {
+      vcalcText <- paste0(
```
juliuspfadt, line 604: For the `"precomputed"` V-matrix type, the generated R code assigns the file path as a bare quoted string:
```r
effectSizeVarianceCovarianceMatrix <- '/path/to/file.csv'
```
A user copying this gets a `character` variable, not a matrix. The assignment should load the file:
```r
effectSizeVarianceCovarianceMatrix <- as.matrix(read.csv("/path/to/file.csv", header = FALSE))
```
The non-precomputed branch (lines 608–614) correctly generates a `vcalc(...)` call that produces an actual matrix.

juliuspfadt:
## Additional findings from re-review (post-update pass)

### R/classicalmetaanalysiscommon.R:1532 — `isTryError(fit)` on list, not `fit[[1]]`

```r
if (is.null(fit) || (length(fit) == 1 && jaspBase::isTryError(fit)) || ...
```

`fit` here is the **list** returned by `.maExtractFit()`. `jaspBase::isTryError` checks `inherits(x, "try-error")`; a list is never of class `try-error`, so this guard always passes (never fires) when the single element is a failed fit. Execution then falls through to `.maMakeTheUltimateForestPlot(fit, options)` which crashes instead of returning cleanly.

Every other structurally identical guard in the same file (lines 775, 934, 1004, 1060, 1122, 1572, 1738, 1867) correctly uses `jaspBase::isTryError(fit[[1]])`. Should be:
```r
if (is.null(fit) || (length(fit) == 1 && jaspBase::isTryError(fit[[1]])) || ...
```

---

### inst/Upgrades.qml — missing `ChangeRename` for `forestPlotAllignLeftPanel` → `forestPlotAlignLeftPanel`

The PR fixes the typo (`Allign` → `Align`) in both QML and R code. However `inst/Upgrades.qml` has no `ChangeRename` entry for this rename. Existing `.jasp` files that stored the alignment preference under the old key `forestPlotAllignLeftPanel` will silently revert to the default alignment on load.

Needed:
```qml
ChangeRename { from: "forestPlotAllignLeftPanel"; to: "forestPlotAlignLeftPanel" }
```

---

### R/effectsizecomputation.R:201, 558 — `"SMDH1"` typo (unchanged context, same as line 992 flagged inline)

Lines 201 and 558 also contain `c("SMD1", "SMDH1")` — same transposition, same effect. All three occurrences should read `c("SMD1", "SMD1H")`. The inline comment on line 992 covers the same root cause.

---

### inst/qml/EffectSizeComputation.qml:638 — `"SMDH1"` in visibility expression

```qml
visible: (designValue === "independentGroups" && ... && (effectSizeValue != "SMD1" && effectSizeValue != "SMDH1")) || ...
```

`"SMDH1"` should be `"SMD1H"`. Because this condition never matches when `effectSizeValue === "SMD1H"`, the control remains visible when it should be hidden (or vice versa) whenever the user selects SMD1H.

juliuspfadt (changes requested):
fix error

FBartos:
> this seems like its not fixed: https://github.com/jasp-stats/jaspMetaAnalysis/pull/436#issuecomment-4705738320
I see a similar error with this
[forestErrorAgain.jasp.zip](https://github.com/user-attachments/files/29098263/forestErrorAgain.jasp.zip)

Once I refresh the file the error disapears, are you sure you are on the new version?

R/classicalmetaanalysiscommon.R
FBartos: the new behavior is correct

R/effectsizeaggregation.R
FBartos: that is intended, the code is mostly illustrative as more processing is done to the file anyways

R/classicalgeneralizedmetaanalysis.R
FBartos: fixed

R/effectsizecomputation.R
FBartos: fixed

R/classicalmetaanalysiscommon.R
FBartos: fixed

inst/Upgrades.qml
```diff
@@ -338,6 +338,81 @@ Upgrades
...
+		ChangeJS
+		{
+			name:		"exportDiagnosticsCoefficientInfluence"
+			jsFunction:	function(options)
+			{
+				return options["diagnosticsCasewiseDiagnosticsExportToDataset"] === true &&
+					options["diagnosticsCasewiseDiagnosticsDifferenceInCoefficients"] === true;
+			}
+		}
+	}
```
juliuspfadt, line 462: The `ClassicalMetaAnalysis` 0.95.4→0.95.5 block adds `ChangeRename { from: "forestPlotAllignLeftPanel"; to: "forestPlotAlignLeftPanel" }` (line 388), but this block for `ClassicalMetaAnalysisMultilevelMultivariate` does not. Both analyses share the `ForestPlotSection.qml` component which contains the alignment option. Without the rename, `.jasp` files created with MAMM in 0.95.4 will silently lose the alignment preference when opened in 0.95.5+.

Add before the closing `}`:
```qml
ChangeRename { from: "forestPlotAllignLeftPanel"; to: "forestPlotAlignLeftPanel" }
```

juliuspfadt:
## Missing `ChangeRename` for alignment fix in other analyses

The typo fix `forestPlotAllignLeftPanel` → `forestPlotAlignLeftPanel` (in `ForestPlotSettings.qml`) requires a `ChangeRename` migration in **every** analysis that used the forest plot before this PR. Otherwise old `.jasp` files silently revert the alignment to the default.

The fix was added to **`ClassicalMetaAnalysis`** (line 388) and separately flagged for **`ClassicalMetaAnalysisMultilevelMultivariate`** (inline comment). But the following analyses also use `ForestPlotSection.qml` and are missing migrations entirely (no 0.95.4→0.95.5 upgrade block exists for them):

| Analysis | Status |
|----------|--------|
| `ClassicalMetaAnalysis` | ✅ has ChangeRename |
| `ClassicalMetaAnalysisMultilevelMultivariate` | ❌ block exists, ChangeRename absent |
| `BayesianMetaAnalysis` | ❌ no 0.95.4→0.95.5 block |
| `BayesianBinomialMetaAnalysis` | ❌ no 0.95.4→0.95.5 block |
| `ClassicalMantelHaenszelPeto` | ❌ no 0.95.4→0.95.5 block |
| `ForestPlot` (standalone) | ❌ no 0.95.4→0.95.5 block |
| `RobustBayesianMetaAnalysis` | ❌ no 0.95.4→0.95.5 block |
| `ClassicalGeneralizedMetaAnalysis` | ✅ new in this PR, no saved files exist |

For analyses with no block, a new `Upgrade` entry is needed:
```qml
Upgrade {
    functionName: "BayesianMetaAnalysis"
    fromVersion:  "0.95.4"
    toVersion:    "0.95.5"

    ChangeRename { from: "forestPlotAllignLeftPanel"; to: "forestPlotAlignLeftPanel" }
}
```
(repeat for each analysis above)

juliuspfadt:
Error is fixed. Can you look at the comments Claude generated?

juliuspfadt (changes requested):

R/forestplotcommon.R
```diff
@@ -0,0 +1,2529 @@
...
+
+# Adapt the incoming fit list into a stable shape before the later section and
+# layout builders start combining subgroup/full-dataset outputs.
+.forestPlotPrepareFitItems             <- function(fit, options) {
+
+  if (options[["subgroup"]] != "" && length(fit) > 1) {
+    fit <- fit[c(2:length(fit), 1)]
+  }
+
+  if (options[["subgroup"]] != "" && length(fit) > 0) {
```
juliuspfadt, line 111: **Bug: `fullDatasetName` incorrectly tags a subgroup fit as the full dataset when `includeFullDatasetInSubgroupAnalysis = FALSE`**

When this option is `FALSE`, `.maExtractFit` removes the `"__fullDataset"` entry before passing the fit list to the forest plot. The incoming list is therefore `[sg1, sg2, sg3]` — no full-dataset entry.

The reorder at line 108 then moves `sg1` to the last position (`[sg2, sg3, sg1]`), and this line sets:
```r
fullDatasetName <- names(fit)[length(fit)]  # → "sg1", not the full dataset
```

Any surviving subgroup named `sg1` then receives `isFullDataset = TRUE` in the item builder below, causing:
- Study rows for `sg1` to be suppressed (if `.forestPlotShouldIncludeStudySection` gates on `isFullDataset`), or
- EMM / model-information sections marked `includeWithFullDataset = FALSE` to be incorrectly included for `sg1`.

**Fix:** The reorder and the `fullDatasetName` assignment assume the first element of the fit list is always `"__fullDataset"`. When `includeFullDatasetInSubgroupAnalysis = FALSE` this invariant does not hold. Guard the block:
```r
if (options[["subgroup"]] != "" && length(fit) > 0 &&
    "__fullDataset" %in% names(fit)) {
  fullDatasetName <- "__fullDataset"
} else {
  fullDatasetName <- NULL
}
```
and adjust the reorder similarly.

R/forestplotcommon.R
```diff
@@ -0,0 +1,2529 @@
...
+    } else {
+      .forestPlotResolveXAxisLabels(xAxis[["breaks"]], options)
+    }
+
+    xRange  <- if (labelsOnly) .forestPlotInverseAxisValues(xAxis[["range"]],  transformationSpec) else xAxis[["range"]]
+    xBreaks <- if (labelsOnly) .forestPlotInverseAxisValues(xAxis[["breaks"]], transformationSpec) else xAxis[["breaks"]]
+  }
+
+  xPadding <- .forestPlotResolveXAxisPadding(plotData, xRange)
+  yRange   <- c(min(yValues, na.rm = TRUE) - .forestPlotRowSize(options), 0)
```
juliuspfadt, line 2043: **Bug: `min(yValues, na.rm = TRUE)` crashes when all subgroup fits fail**

When multiple subgroups are active and every fit is a `try-error`, the guard in `.maUltimateForestPlot` only catches the single-fit-errored case (`length(fit) == 1 && isTryError(fit[[1]])`). With two or more all-failing subgroups the guard is skipped.

Inside `.forestPlotPrepareFitItems` the error filter (line 116) removes all try-errors, producing an empty `fitItems` list. No rows are added to the layout, so every data frame in `plotData` is NULL and:
```r
yValues <- c(...)  # numeric(0)
yRange  <- c(min(yValues, na.rm = TRUE) - ..., 0)  # c(Inf, 0)
```
`min(numeric(0), na.rm = TRUE)` returns `Inf` (with a warning). The resulting `yRange = c(Inf, 0)` is passed to `coord_cartesian(ylim = ...)` which throws an error about infinite limits. The `try()` in the caller catches it and the plot shows a red error banner instead of rendering.

**Fix options:**
1. Add a guard before this line: `if (length(yValues) == 0) return(...)` with a sensible default range.
2. Or broaden the guard in `.maUltimateForestPlot`: `all(vapply(fit, jaspBase::isTryError, logical(1)))`.

juliuspfadt (approved):
This looks good, and I dont think the new comments from Claude are a show-stopper. You can fix or not, whatever you feel like.

R/forestplotcommon.R
FBartos: not correct, the forest plot handles this in the constructor so full dataset is always included:

```
  # the full data set fit is always needed for subgroup analyses
  # there are forest plot specific settings
  options[["includeFullDatasetInSubgroupAnalysis"]] <- TRUE
  fit <- .maExtractFit(jaspResults, options)
```

R/forestplotcommon.R
FBartos: fixed

PR #443, opened 2026-06-22
Handle unavailable diagnostics for location-scale model

FBartos:
Make diagnostics/export robust for location–scale (heterogeneity metaregression) models and for fit types that don't support certain outputs.


NEWS.md
```diff
@@ -1,41 +1,37 @@
...
+> * **Adding New Changes (For Contributors):** All new commits should be logged at the very top of the file under the `# jaspMetaAnalysis (development version)` header. Place your bullet point under the appropriate category (`## Added`, `## Changed`, `## Fixed`, etc.).
+> * **Issue References:** Please reference the relevant GitHub Issue (if any) at the end of your line (e.g., `([Issue #19](https://github.com/jasp-stats/jaspMetaAnalysis/issues/19))`).
+> * **Format Categories:**
+>   * **Added:** New analyses, features, options, or output.
+>   * **Changed:** Updates to existing analyses, defaults, dependencies, or output.
+>   * **Fixed:** Bug fixes in analyses, plots, tables, help, or QML layouts.
+>   * **Deprecated / Removed:** Outdated analyses, options, or legacy code.
 
 ---
+# jaspMetaAnalysis 0.95.6
```
juliuspfadt: Version mismatch: DESCRIPTION bumps to **0.96.6** but this heading says **0.95.6** (0.96 vs 0.95). Looks like a typo — should be `# jaspMetaAnalysis 0.96.6`.

R/classicalmetaanalysiscommon.R
```diff
@@ -2123,7 +2200,7 @@ ClassicalMetaAnalysisCommon <- function(jaspResults, dataset, options, ...) {
 
   return(columns)
 }
-.maExportVectorFromFitList                <- function(dataset, fit, exportFunction) {
+.maExportVectorFromFitList                <- function(dataset, fit, exportFunction, isAvailable = NULL) {
```
juliuspfadt, line 2203: This change also drops the old `if (!hasValues) return(NULL)` further down, so an all-NA column is now *always* exported when the option is on — including for non-`rma.ls` fits where no fit supports the type. Consistent with the PR's "keep the column, fill NA" approach, just confirming that global behavior change is intended (e.g. conditional residuals), not only for location-scale.

R/classicalmetaanalysiscommon.R
```diff
@@ -1943,34 +1943,106 @@ ClassicalMetaAnalysisCommon <- function(jaspResults, dataset, options, ...) {
 .maExportDiagnosticsCoefficientInfluence  <- function(options) {
   return(isTRUE(options[["exportDiagnosticsCoefficientInfluence"]]))
 }
+.maExportUnavailableDiagnosticsColumns    <- function(jaspResults, dataset, options) {
```
juliuspfadt, line 1946: No regression test accompanies this new branch. A `testAnalysis("ClassicalMetaAnalysis")` case with `heterogeneityModelTerms` set plus diagnostics/export enabled would lock in the fix and guard against these models regressing back to a hard error.

juliuspfadt (commented):
Reviewed the location-scale diagnostics/export handling. Logic verified against a live `metafor` session (location-scale fit is `c("rma.ls","rma.uni","rma")`; `rstandard` works, `rstudent` errors), so the `.maExportFitSupports*` predicates are correct. Three small points inline.

R/classicalmetaanalysiscommon.R
FBartos: yes, that is the inteded change

R/classicalmetaanalysiscommon.R
FBartos: would be overkill

NEWS.md
FBartos: fixed

juliuspfadt:
any pointers as to how I can test this?

FBartos:
sorry, need to unzip this:
[diag-errors.zip](https://github.com/user-attachments/files/29236396/diag-errors.zip)

juliuspfadt (approved):

PR #463, opened 2026-08-14
meta-analytic SEM and SEM-based meta-analysis fixes

FBartos:
- Restore OpenMx globals (`mxOptions`, `mxByrow`, SLSQP) *before* `library(metaSEM)`. JASP `restoreOptions()` NULLs those options while OpenMx stays loaded, so metaSEM `.onLoad` crashed with missing `"Default optimizer"`. Keep `library(metaSEM)` after the restore so OpenMx is on the search path for `mxRun`/`osmasem2`.
- Fix MASEM follow-ons: convergence-table footnotes used `modelFitTable`; path diagrams `return()`’d on a NULL fit and skipped later models.
- SEM-based MA only: `lavaan2RAM(std.lv = FALSE)` (no UI toggle; `f ~~ 1*f` in syntax if needed). `.masemGetObservedVariables()` returns encoded `syntax$columns` only — not `data.` definition-variable labels, which made `obs.variables` disagree with the model (Jaramillo).


FBartos:
accidentally pushed the changes directly to master before

PR #467, opened 2026-08-19
Fix ordinal moderator casting

### LINKED ISSUE #4481: [Bug]: Meta-analysis modules in multiple JASP versions
### JASP Version

0.98.1

### Commit ID

_No response_

### JASP Module

Meta Analysis

### What analysis are you seeing the problem on?

_No response_

### What OS are you seeing the problem on?

macOS Silicon

### Bug Description

I completed my bachelor’s thesis at the UvA in January 2026 using JASP version 0.95.4. 

I conducted a multilevel meta-analysis with the following specification:

- Effect size: Cohens_D
- Standard error: Cohens_D_SE
- Nested levels: Study_ID and Effect_size_ID
- Moderator: a three-category instructional-method variable
- Estimation method: Restricted Maximum Likelihood (REML)
- Fixed-effects test: t

The analysis worked correctly at the time and produced a pooled effect size of approximately 0.34, which was plausible given the observed effect sizes.

I am now trying to reproduce the analysis using exactly the same dataset and JASP version 0.95.4. When the Meta-Analysis module recalculates the analysis, it terminates without producing any output and displays the following message:

Image

I then recreated the same analysis using JASP version 0.98.1. In this version, the analysis runs, but it produces results that appear impossible given the dataset. The Cohen’s d values range from −0.46 to 1.73, but JASP reports a pooled effect size of 8.16. By comparison, the original analysis conducted for my bachelor’s thesis produced a pooled effect size of approximately 0.34.

I am reporting this because the same dataset and equivalent model specification either terminate unexpectedly in JASP 0.95.4 or produce implausible results in JASP 0.98.1.

### Expected Behaviour

Using the same dataset and model specification should reproduce the original multilevel meta-analysis, or at least produce reasonably similar results across JASP versions.

In JASP 0.95.4, the analysis should run without terminating and should reproduce the original pooled effect size of approximately 0.34.

In JASP 0.98.1, the pooled effect size should be consistent with the scale and distribution of the observed Cohen’s d values. A pooled effect size of approximately 8.16 is not plausible when the individual effect sizes range from approximately −0.46 to 1.73.

### JASP version 0.95.4

1. Open the original dataset or the JASP file used for the bachelor’s thesis.
2. Select the Classical Multilevel/Multivariate Meta-Analysis.
3. Assign Cohens_D as the effect-size variable.
4. Assign Cohens_D_SE as the effect-size standard-error variable.
5. Specify Study_ID and Effect_size_ID as the nested levels.
6. Add the three-category instructional-method variable as the moderator.
7. Select Restricted Maximum Likelihood and the t test.
8. Wait for JASP to run the analysis.
9. The analysis terminates and displays: 

> This analysis terminated unexpectedly.
> 
> Error in apply(dataset[, omitOnVariables], 1, function(x) anyNA(x)): dim(X) must have a positive length
> 
> Stack trace
> tryCatchOne(expr, names, parentenv, handlers[[1]])
> 
> doTryCatch(return(expr), name, parentenv, handler)
> 
> withCallingHandlers(expr = analysis(jaspResults = jaspResults, dataset = dataset, options = options), error = .addStackTrace)
> 
> analysis(jaspResults = jaspResults, dataset = dataset, options = options)
> 
> .mammCheckData(dataset, options)
> 
> apply(dataset[, omitOnVariables], 1, function(x) anyNA(x))
> 
> stop('dim(X) must have a positive length')
> 
> To receive assistance with this problem, please report the message above at: https://jasp-stats.org/bug-reports

### JASP version 0.98.1

1. Recreate the same multilevel meta-analysis using the variables and settings listed above.
2. Wait for JASP to run the analysis.
3. The analysis completes but reports a pooled effect size of 8.16 instead of the original value of 0.34.

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

- [x] I have included a screenshot showcasing the issue, if possible.
- [ ] I have included an encrypted JASP file (zipped) or data file that causes the crash/bug, if applicable.
- [x] I have accurately described the bug, and steps to reproduce it.

FBartos:
Add allowedColumns: ["scale","nominal"] to variable selector controls to restrict selectable variable types to scale or nominal. Prevents invalid type selection in BubblePlot, ForestPlot and Estimated Marginal Means UI components.

fixes: https://github.com/jasp-stats/jasp-issues/issues/4481

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: jaspMetaAnalysis
 Type: Package
 Title: Meta-Analysis Module for JASP
-Version: 0.97.2
+Version: 0.97.3
 Date: 2022-10-05
 Author: JASP Team
 Website: jasp-stats.org
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,10 @@
 >   * **Deprecated / Removed:** Outdated analyses, options, or legacy code.
 
 ---
+# jaspMetaAnalysis 0.97.3
+## Fixed
+* Predictor type casts not being preserved in estimated marginal means and prediction plots.
+
 # jaspMetaAnalysis 0.97.2
 ## Fixed
 * Meta-analytic SEM failing to load because JASP wipes OpenMx `mxOptions` before `metaSEM` `.onLoad`.
--- a/inst/qml/qml_components/BubblePlot.qml
+++ b/inst/qml/qml_components/BubblePlot.qml
@@ -43,6 +43,7 @@ Section
 			id:				bubblePlotSelectedVariable
 			title:			qsTr("Selected Variable")
 			singleVariable:	true
+			allowedColumns:	["scale", "nominal"]
 			allowTypeChange:false
 			info: qsTr("Variable to be visualized on the x-axis.")
 		}
@@ -52,6 +53,7 @@ Section
 			name:			"bubblePlotSeparateLines"
 			id:				bubblePlotSeparateLines
 			title:			qsTr("Separate Lines")
+			allowedColumns:	["scale", "nominal"]
 			allowTypeChange:false
 			info: qsTr("Variable(s) according to which predictions are split across different lines.")
 		}
@@ -61,6 +63,7 @@ Section
 			name:			"bubblePlotSeparatePlots"
 			id:				bubblePlotSeparatePlots
 			title:			qsTr("Separate Plots")
+			allowedColumns:	["scale", "nominal"]
 			allowTypeChange:false
 			info: qsTr("Variable(s) according to which predictions are split across different plots.")
 		}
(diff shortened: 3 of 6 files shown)


sisyphus-jasp:
🚀 I’ve accepted this review request. The native worker will inspect the current pull request and run independent tests before publishing the review.

sisyphus-jasp:
🔎 Sisyphus finished the coding pass and started an independent review.

Automation job: `7da007dc-ee78-496a-ad43-5f8cbd23cbe3`

sisyphus-jasp (changes requested):
Requested changes: add ordinal moderator cast regression coverage; focused-tests and QML lint passed.
### MEDIUM: Add regression coverage for moderator casts
Location: `inst/qml/qml_components/BubblePlot.qml:46`
Could you add a focused regression fixture/test that starts with an ordinal data column cast to nominal or scale, assigns it to the EMM, bubble-plot, and forest-plot selectors, reloads the saved options, and verifies the cast and resulting outputs? The PR changes only QML and adds no regression test; the closest existing all-destination case in the GLMM test was skipped by the focused check and its fixture uses native nominal/scale columns, so it cannot detect the reported cast-loss failure.

Automation job: `7da007dc-ee78-496a-ad43-5f8cbd23cbe3`

juliuspfadt:
build builder bot

juliuspfadt:
whats with the tests? I mean, they do also fail on master, but still, some drift in the numbers?

juliuspfadt:
also builder bot seems to have issue with OpenMx? @RensDofferhoff?

juliuspfadt (approved):
Fixed.

